PR #1, opened 2016-09-28
draft of proposal for autoloading and settings

galuszkak:


galuszkak:
Some review before end of hackathon will be helpful @prusnak ;), so I can say if I contributed anything on this hackathon ;)

spudowiar:
Can we have app dependencies? (e.g. OpenPGP depends on CCID)

jpochyla:
We were planning something similar, but i'd like to avoid importing the apps automatically at the moment. It boils down to simply `__import__`ing the configured modules from `settings.config.APPS`, and it seems like an overkill for now, and replacing the same amount of code with the same amount of configuration.

Nevertheless, a config or settings compile-time file seems like a good idea, i'll merge in some unfinished stuff later and merge it in!

galuszkak:
@jpochyla idea is to start moving configuration things from different apps into one place to easily change something without going into source code. 

This is more convenient later when for user/developer it's just 3rd party app downloaded from PYPI and just adding setting that are needed for that app.

prusnak:
Thanks for the concept idea. We'll rethink the idea of pluggable/configurable apps later. Now development is in rapid mode and we are breaking a lot of things and it's easier not to include extra stuff.

Issue #8, opened 2017-08-28
docs: info on display for dev kit


mcudev:
I want to update my dev kit to support TREZOR 2 development (https://mcudev.github.io/trezor-dev-kit/index.html).

I need a display.

It looks like the basic requirements are:

- 240x240 pixel
- TFT LCD 18-bit color screen
- ILI9341V or ST7789V driver
- I2C capacitive touch

Do you have any recommendations on a module to use (preferably on a breakout board with headers)?

I see some 240x320 modules, but not 240x240. Is 240x320 what is actually being used?

For example, this looks close to spec: https://www.adafruit.com/product/2090

prusnak:
Our display is indeed 240x240, but it uses 240x320 pixel buffer internally. Also we currently use ST7789V as the driver. Data bus is 8-bit parallel.

Your selection is fine, though. ILI9341V driver was our original choice, it's not too different from ST7789V. Also nice that your choice has SD-card as TREZOR2 will contain it too.

mcudev:
Thanks for the info!
Indeed, the SD-card being included on that adafruit board is extra convenient.
Quick question, are you getting touch input over the same 8-bit parallel bus? Or, is only screen output over that bus, and touch input comes over I2C?
I quick-skimmed the code and it looked like I2C, but could be wrong.

prusnak:
Correct, touch comes over I2C.

Also, if you think it would be nice to have all this info in one place, I would love to accept a pull request which adds `hardware.md` into `docs/` directory or something similar!

(Btw, I'll close this issue, but feel free to ask more questions, I'll answer twem)

mcudev:
Pull request created https://github.com/trezor/trezor-core/pull/9.

I assumed that doc should be targeted to v2 only.
Do you have any interest in pulling the existing dev kit stuff that I have into one of your repos (https://mcudev.github.io/trezor-dev-kit/index.html)?
I'd be happy to contribute it. It would probably get better use/exposure being in one of your repos.

Another question:
I was looking closer at the adafruit board.
It looks like the microSD is not compatible because that board is setup to use SPI and your code uses a 4-bit data bus for microSD.
Am I reading that correctly?

mcudev:
Nevermind that last question. I'm sure the code is using 4-bit SDIO mode through the SDIO peripheral and not SPI. I think I'll use a cheap separate SDIO breakout and am still looking for a ST7789V display module with cap touch that's easy to use.

PR #9, opened 2017-08-29
docs: add reference hardware and dev kit info

mcudev:
This is a rough draft (not ready to be merged) to get the discussion going.
I'm trying to piece together what is needed from the code, and searching Google.
For example, there are some TODOs with questions.

Referencing https://github.com/trezor/trezor-core/issues/8

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,3 +4,4 @@
 * [TREZOR Core Bootloader](bootloader.md)
 * [TREZOR Core Image Format](toif.md)
 * [TREZOR Core Memory Layout](memory.md)
+* [TREZOR Core Hardware](hardware.md)
--- a/docs/hardware.md
+++ b/docs/hardware.md
@@ -1,0 +1,102 @@
+# TREZOR Core Hardware
+
+## TREZOR v2 Open Source Hardware Reference Documentation
+
+### Photo Front
+TODO
+
+### Photo Back
+TODO
+
+### Bill of Materials / BOM
+TODO
+
+### Eagle Schematic
+TODO
+
+### Eagle Board
+TODO
+
+## Developer Kit
+
+* 1 x [Waveshare Core405R Dev Board](http://www.waveshare.com/core405r.htm)
+* 1 x USB Cable Type A Plug/Male to Type Mini-B Plug/Male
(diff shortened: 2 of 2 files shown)


prusnak:
Thanks!

prusnak:
We are using LQFP100 variant of STM32F405, because we need FSMC controller to push 8-bit parallel into the display.

mcudev:
Ah, I was wondering that when I saw it was not supported on the STM32F405RG in the datasheet. I was hoping it was a mistake.

mcudev:
I was wondering if you tried SPI for the SD card and found it unacceptable? Using SPI instead makes it more accessible for hobbyist use and would allow the design to go back to the LQFP64.

prusnak:
We used the reference design from pyboard and this used SDIO, but I think SPI should be also fine.

mcudev:
The pyboard uses a STM32F405RGT6 though. How could it use SDIO without the FSMC memory controller? FWIW, breakout boards with a display, touch, and SPI microSD are easier to get.

prusnak:
We use FSMC for display, SDIO for SD card.

mcudev:
Doh, you're right, I mixed that up. The displays are commonly driven with SPI on the breakouts too. Any chance that SPI could be used for the display, or is that too slow?

prusnak:
The display we picked does not have SPI, that's why we use FSMC. I think you could be fine with SPI as well. The change needed to support both should be really trivial in code.

mcudev:
Interesting. Thanks for the help! I'll look into it.

prusnak:
I would say: try getting a breakout board with SPI display and microSD and we'll update the code to work on both platforms (SPI+SPI and FSMC+SDIO) together. Should not be a big issue at all.

mcudev:
Wow, cool, will do.

mcudev:
I found some parts that may match closer to the existing code. I'll update you after I try them.

Which version of the pyboard reference design did you use (I'm guessing 1.0)? The differences are listed at the bottom of https://store.micropython.org/#/features.

It looks like the code is configured for an 8 MHz high-speed external crystal. That's a good common frequency on available parts -- are you sticking with that?

prusnak:
Pyboard v1.1 was used as the reference. But we use 8MHz crystal.

PR #10, opened 2017-09-05
fix minor compile time bug and update docs

mcudev:
the compile time bug only occurs when compiling with DISPLAY_ILI9341V for that type of display.
the rest of the changes are comment adds and documentation updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ cd trezor-core
 sudo dpkg --add-architecture i386
 sudo apt-get update
 sudo apt-get install scons libsdl2-dev:i386 libsdl2-image-dev:i386 gcc-multilib
+pip install ed25519 pyblake2
 make build_unix
 ```
 
--- a/docs/hardware.md
+++ b/docs/hardware.md
@@ -19,10 +19,9 @@ TODO
 
 ## Developer Kit
 
-* 1 x [Waveshare Core405R Dev Board](http://www.waveshare.com/core405r.htm)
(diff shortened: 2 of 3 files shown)


prusnak:
Thank you!

Btw, I see you changed the dev board to Discovery Board from STM32.

This one is also quite interesting - http://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-eval-tools/stm32-mcu-eval-tools/stm32-mcu-discovery-kits/32f429idiscovery.html - $10 more than your selection, but it already contains a ILI9341 display. Not sure how it is wired (most probably SPI, not 8-bit parallel) or if it has PWM backlight.

mcudev:
Interesting, thanks for sharing the info. It looks like the 32F429IDISCOVERY has its display pins mapped to different GPIO ports/pins than we need though. In that respect separate display breakouts are more flexible. Also looks like a resistive touch panel.

I may change the dev board again. I've got some more stuff in the pipeline to try. One annoying thing about the current board is that it won't power through the micro-USB port. So you need both USB's plugged-in (which is kind of dorky).

mcudev:
I just tried firing up the 32F429IDISCOVERY and found that the MCU interface mode is set with some tiny resistors to "4-wire 8-bit serial I, SDA:In/Out". Like you thought might be the case, I couldn't find a PWM pin either. If it has one, it's disconnected. On to the next one.

PR #11, opened 2017-09-07
better support ILI9341 testing, make tearing effect synch optional, minor fixes

mcudev:
My current test display is ILI9341 and doesn't have the tearing effect line connected.
The build changes scratch that itch while defaulting to what you had already.
Example usage:
`make build_boardloader TEARING_EFFECT=1 DISPLAY_ILI9341V=1`

I removed the red flash on fatal error to avoid possibly triggering epileptic seizures for people.

Code review question:
Does the ST7789V really use RGB while the ILI9341V uses BGR? I tested with the ILI9341 and BGR works for it. I don't have my ST7789V display yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,14 @@ UNIX_PORT_OPTS ?= TREZOR_X86=1
 endif
 CROSS_PORT_OPTS ?= MICROPY_FORCE_32BIT=1
 
+ifeq ($(DISPLAY_ILI9341V), 1)
+CFLAGS += -DDISPLAY_ILI9341V=1
+CFLAGS += -DDISPLAY_ST7789V=0
+endif
+
+ifeq ($(DISPLAY_VSYNC), 0)
+CFLAGS += -DDISPLAY_VSYNC=0
+endif
 
 ## help commands:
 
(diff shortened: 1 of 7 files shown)


mcudev:
CI build failures are unrelated to these changes.

prusnak:
Does the ST7789V really use RGB while the ILI9341V uses BGR?

IIRC ST7789V can use both. I had some problems with my ILI9341 that it did not support RGB correctly, just BGR. So that's why we use RGB for ST7789V and BGR for ILI9341.

If RGB works for your ILI9341, i suggest to use RGB for both. I will wait for you to try it out, and if it works, please change your commit to use RGB everywhere.

Also please rename TEARING_EFFECT to DISPLAY_VSYNC and change the logic of it (so DISLAY_VSYNC is 1 by default).

mcudev:
Updated, thanks.

I was asking because my ILI9341V also seems to only work with BGR. I was wondering if we should use BGR only in `display_set_orientation` to reduce code? Depends on how the ST7789V works. I'll just try it when I get my ST7789V and compact the code later if I can.

prusnak:
Thanks! Merged!

PR #13, opened 2017-09-08
display related updates, flashing++

mcudev:
1) Smoothed out the display (re)initializing and clearing to reduce flashing during startup and clear the whole display for 240x320 displays.
Also, removing the `display_init` from the python binding fixes some sporadic artifacts on the trezor.io/start message (the initial screen after the 1 second green SL logo).

With these changes, there is only one white flash at the very beginning, before the display is initialized. That may be solvable by hardware design on the backight pin (just a guess, I did not try this).

Tested with:
```
make build_boardloader DISPLAY_VSYNC=0 DISPLAY_ILI9341V=1
make build_bootloader DISPLAY_VSYNC=0 DISPLAY_ILI9341V=1
make vendorheader
make build_firmware DISPLAY_VSYNC=0 DISPLAY_ILI9341V=1
```

2) Some function brace cleanup while I was touching the files.

### DIFF
--- a/embed/boardloader/main.c
+++ b/embed/boardloader/main.c
@@ -146,9 +146,6 @@ int main(void)
         __fatal_error("sdcard_init", __FILE__, __LINE__, __FUNCTION__);
     }
 
-    display_clear();
-    display_backlight(255);
-
     display_printf("TREZOR Boardloader %d.%d.%d.%d\n", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, VERSION_BUILD);
     display_printf("==================\n");
     display_printf("starting boardloader\n");
--- a/embed/bootloader/main.c
+++ b/embed/bootloader/main.c
@@ -83,7 +83,6 @@ void check_and_jump(void)
 
         display_vendor(vhdr.vimg, (const char *)vhdr.vstr, vhdr.vstr_len, hdr.version);
         HAL_Delay(1000); // TODO: remove?
-        display_printf("JUMP!\n");
         jump_to(FIRMWARE_START + vhdr.hdrlen + HEADER_SIZE);
 
     } else {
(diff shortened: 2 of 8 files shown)


mcudev:
oops, don't pull this yet, i didn't see you made some commits recently. let me merge those in.

mcudev:
Should be good to merge now. I also added the hardware doc update for the PWM pin, PA7.

mcudev:
I found a bug with this. Don't merge yet. I didn't see that `display_init` sets up `TIM1_Handle`, which is external to the function. I'll have to account for that.

mcudev:
Works now. I tested up through the 5 second homescreen fade.

prusnak:
Please don't mix formatting (whitespace) changes and functional changes in one commit.

Also, can we rename `lcd_pwm_timer_init` to `display_pwm_init`?

mcudev:
Sure, updated. Tested with `make vendorheader build DISPLAY_VSYNC=0 DISPLAY_ILI9341V=1`

mcudev:
I also rebased this onto master today. I think it's still good.

In addition to these changes, if you can keep `PA7` low until `display_pwm_init` runs for the first time, then you can avoid the initial white flash/flicker too.
But, I think you'd have to add a switch that defaulted to low and could be controlled by a GPIO, or an external pull-down, maybe (the GPIO's are floating inputs at reset).

prusnak:
Is that intentional that `display_init` is never called?

mcudev:
It's called from the boardloader's main in embed/boardloader/main.c. Other than that, yes. It's the re-initializing over and over again that I was trying to avoid.

mcudev:
just rebased onto master again to pull in the latest changes

prusnak:
> It's called from the boardloader's main in embed/boardloader/main.c

makes sense, I missed that

prusnak:
Thanks!

prusnak:
JFYI: it is now possible to compile firmware in standalone mode (to be flashed to location `0800000` and skipping the boardloader/bootloader completely) - I added a quick hack in https://github.com/trezor/trezor-core/commit/cfe9c73c7847b84b2afc5be0a3ea58e127e4db60 to detect this and call `display_init` in this case.

We might want to come up with nicer solution later, but I think it's OK for now.

mcudev:
OK, thanks for the heads-up.

PR #15, opened 2017-09-10
hardfault handler update

mcudev:
I noticed this while working on the startup code. To me, it seems unsafe to print memory to the display when that could potentially include sensitive things. What do you think?


prusnak:
Right, this was useful during developing the board, but not a good idea to have this included now.

PR #16, opened 2017-09-18
propose re-mapping sd card detect pin from PC13 to PC7

mcudev:
@prusnak This PR proposes re-mapping the microSD Socket SD_CARDDETECT pin from PC13 to PC7.
I found a dev board that has a microSD socket that supports SDIO mode.
Hopefully I'm still early enough in the process that the reference hardware design can still be changed to accomodate this.
Example dev board: http://www.ebay.com/itm/STM32F407VGT6-ARM-Cortex-M4-32bit-MCU-Core-Development-Board-STM32F4Discovery-/182450698403


prusnak:
Sorry, we cannot change the PIN anymore. The hardware for T2 is fixed already. :-(

mcudev:
Thanks for considering. It turns out that this dev board has poor quality / mechanical design anyways. The USB connector popped off after only a few days of light use.

PR #17, opened 2017-09-27
docs: hardware updates

mcudev:
The most important change here is syncing up the LCD_PWM PB13 to PA7 change.
Also added some more specifics.
By the way, what does SBU stand for? embed/trezorhal/sbu.h


prusnak:
T2 has USB-C port. SBU pins are "SideBand Use" - see https://en.wikipedia.org/wiki/USB-C

We decided to wire these to MCU for future use.

mcudev:
oh, cool! thanks!

Issue #12, opened 2017-09-07
update startup code


prusnak:
to erase all RAM similarly to reset_handler here: https://github.com/trezor/trezor-mcu/blob/master/startup.s

mcudev:
Looks like you might have to add a call to `SystemInit` right before `main` and do all the vector table stuff.
This also depends on the linker script, so probably changes there too.
Do you intend to try to push this upstream to micropython, or just keep it as a patch in https://github.com/trezor/micropython, or something else?

prusnak:
We'll see in the end if it makes sense to upstream that or not. In general we want to upstream most of the stuff, to keep our patchset small.

mcudev:
Makes sense. I've got some work on this. Kind of at the "trying things" stage. The hardest part so far has been reconciling how to deal with all the different sections (due to -fdata-sections -ffunction-sections). There is so much linker input that ends up getting garbage collected (hundreds of sections). I'm trying various approaches to see what ends up clean.

mcudev:
@prusnak I've got a start on this at https://github.com/mcudev/trezor-core/tree/boardloader-update

I've got some more ideas on how to improve it. This is a good start to look at though to see if you like the general direction.

prusnak:
startup code for boardloader was updated via https://github.com/trezor/trezor-core/pull/18

left to do is bootloader and firmware

keeping this open for tracking purposes

mcudev:
Related code review item:
The `jump_to` re-points the exception handlers before the C code that comprises them have been given a good environment to run (that's one of the main purposes of the `reset_handler` that runs after the `jump_to`). So, the `jump_to` probably needs to disable interrupts before the VTOR switch and then the `reset_handler`s need to re-enable interrupts.

For example, the boardloader has three non-default exception handlers configured: PendSV (vector table offset=0x38), SysTick (0x3c), and SDIO (0x104). The SysTick  is definitely configured (by the HAL) and taking interrupts when the code stages are switched.

mcudev:
I was wondering, "What is the point of having a reset_handler for each later stage?". Below is what I came up with. Am I missing anything important?:

 * to setup .data and .bss for the new stage's c code
 * give the new stage of code a fresh/clean/empty stack (but registers etc... are not reset as is done during system reset) (but it could do this itself anyways)
 * give the new stage of code the ability to replace exception handlers (interrupt vectors) (but it could do this itself anyways)
[...]

PR #18, opened 2017-09-28
issue 12: startup related changes

mcudev:
@prusnak Here's some work towards https://github.com/trezor/trezor-core/issues/12. It's ok, but, I'm not sure I totally like it. I have some other versions that don't depend on the HAL and are way smaller and incorporate a lot of other review notes that I haven't mentioned yet. Maybe take a glance and let me know your reaction / first opinion, please.

I think the reason why I removed the exception handlers in this commit (from a few weeks ago IIRC) was because the code they run is not reliable during those exceptions anyways.


prusnak:
Generally, I really like the direction this is going.

We are in the process of finalizing boardloader (by Monday), so you are more than welcome to share other stuff that needs review.

Just a really small nitpick, can we keep `FLASH`/`CCMRAM`/`RAM` uppercase and `rx`/`wal` lowercase to stay consistent with micropython?

Also, now that we dropped `__fatal_error` from all handlers, can we at least indicate we are in the `default_handler` (simple pattern using display PWM comes to mind). What do you think>

prusnak:
Also `NMI_Handler`, `SVC_Handler` and `DebugMon_Handler` were empty, but now they will result in infinite loop. Is that intended?

mcudev:
> Just a really small nitpick, can we keep FLASH/CCMRAM/RAM uppercase and rx/wal lowercase to stay consistent with micropython?

Done, just rebased and force pushed.

> Also, now that we dropped __fatal_error from all handlers, can we at least indicate we are in the default_handler (simple pattern using display PWM LED comes to mind).

Maybe in things after the boardloader. But early in the boardloader, no. IIRC, I came to this conclusion while playing around glitching my clock (while testing some stuff I had done around the clock security system). On my dev board, it was really easy to trigger the CCS by touching a specific pin (I guess my finger had enough capacitance to throw off the HSE).

> Also NMI_Handler, SVC_Handler and DebugMon_Handler were empty, but now they will result in infinite loop. Is that intended?

Yes, I figured that was one safe way to fail and is easy to review :) Definitely the case for NMI because that's where the CSS fails to.

> We are in the process of finalizing boardloader (by Monday), so you are more than welcome to share other stuff that needs review.

OK, I'll need to dig out my notes and other code tonight and refresh my thinking (it's been like a month). Generally, things I had on my mind were:
 * Setup clocks (clock security system too) and RNG before wiping memory?
 * Any other peripheral local memories that should/could be cleared? e.g.- display, usb, touch sensor, fsmc, sdio?
 * port over mature `memory_protect` and RNG using code from TREZORv1
 * glitching protection and checks (I've read some of the public literature about v1 attacks)
 * there's an awful lot that goes into linking the boardloader right now that is unnecessary; could definitely be cleaned up; want the boardloader very small and well understood because that's what gets write protected.
 * and more, but I forget right now...

prusnak:
> Done, just rebased and force pushed.

Thank you, just merged!

> On my dev board, it was really easy to trigger the CSS by touching a specific pin (I guess my finger had enough capacitance to throw off the HSE).

That's how I was checking whether CSS works on T1  :-)

> OK, I'll need to dig out my notes and other code

Please do, would love to see these!

PR #20, opened 2017-09-30
Makefile GDB updates

mcudev:
I find these helpful.
I was thinking about leaving the original `gdb` rule alone, in-case you like how it works for firmware debugging (or making `gdb_firmware` the new `gdb`).
I tend to prefer the automatic `monitor reset halt` and lack of confirmation prompts as well as separate targets for different symbols.


prusnak:
Spot on! Thanks!

PR #22, opened 2017-09-30
flash_otp_write: uninitialized variable

mcudev:
I was compiling a test with `CCFLAGS='-Og '` instead of `CCFLAGS='-Os '` and came across this:
```
embed/trezorhal/flash.c: In function 'flash_otp_write':
embed/trezorhal/flash.c:123:12: error: 'ret' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     return ret;
```


prusnak:
Thanks!

PR #24, opened 2017-10-04
rng updates

mcudev:
This is a follow-up (it builds on) to https://github.com/trezor/trezor-core/pull/21 (that is, this PR is smaller when diffed against #21).

This enables the RNG during `SystemInit` and uses its output to make a first pass over memory. It also changes to follow guidance from the reference manual RM0090 section 24.3.1 about the FIPS continuous random number generator test (as was done in TREZOR v1).


prusnak:
Looks nice - just two comments wrt using CMSIS constants instead of pure values.

mcudev:
I don't see your comments. Did you submit them?

prusnak:
Just now, sorry :-/

prusnak:
Thanks!

PR #25, opened 2017-10-05
display: minor backlight handling

mcudev:
Minor updates for your last commit 6693d61aa072d831d77ffad02a66892873ab1220

1) turn on back light to see `copy_sdcard` output

2) use `display_backlight` instead of `display_set_backlight` to also set `DISPLAY_BACKLIGHT` in embed/extmod/modtrezorui/display.c

I like that you turn the back light off earlier now. Good idea. At one point, I was thinking of only using the display in the boardloader for the SD card update messages. But, then I decided to leave the display init in the normal boot flow so that the display memory got cleared.

### DIFF
--- a/embed/boardloader/main.c
+++ b/embed/boardloader/main.c
@@ -47,6 +47,7 @@ static void progress_callback(uint16_t val) {
 
 bool copy_sdcard(void)
 {
+    display_backlight(255);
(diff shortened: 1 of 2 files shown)


prusnak:
Thanks!

PR #26, opened 2017-10-08
boardloader and bootloader: vector table, jump_to, etc...

mcudev:
This PR is a progression of commits that are prep work for https://github.com/trezor/trezor-core/issues/12 and also adds memory clearing upon leaving a stage of code, like was done by v1 (performed within `jump_to`).

I left the commits unsquashed to aid reviewing (it may help to be able to diff between commits to see the progression).

For the vector table updates, this command helps for reviewing:
`arm-none-eabi-objdump -D -z -j .vector_table build/boardloader/boardloader.elf`

Some of this will get cleaner as the work for https://github.com/trezor/trezor-core/issues/12 gets done too.


prusnak:
Looks great! Keeping the commit unsquashed. Thank you!

prusnak:
Also moving stuff from embed/common to embed/trezorhal in 7bbfba1ac0b276cf45dbf667a64d7ebd74650320

PR #27, opened 2017-10-09
boardloader: simplify linker script

mcudev:
More prep work that makes the bootloader and firmware portion of #12 much easier.
I have the bootloader and firmware updates too. I'll push them up later, after more testing.

1) The changes like this: `KEEP(*(.text))` to `*(.text*)`
   These changes fully embrace the link time garbage colector and remove the reliance on details of orphan section processing.
   The script is much more intuitive now. I'm glad that I got past the mental block that made me want to use the `KEEP`s.

2) Sets the FLASH memory region to 32K to act as a build time assertion on the code size

3) Other trivial general cleanup/simplifying changes

### DIFF
--- a/embed/boardloader/memory.ld
+++ b/embed/boardloader/memory.ld
@@ -3,14 +3,12 @@
 ENTRY(reset_handler)
 
 MEMORY {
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks!

PR #32, opened 2017-10-09
boardloader: memset_reg_wrap

mcudev:
I had this idea floating around... instead of just writing an unpredictable value, also write the range starting from an unpredictable place. I just imagine this making a power analysis type side channel more difficult. It's probably already sufficiently difficult. But, I figured that I'd throw this out there anyways.

Basically, this picks a random word-aligned address within the range to start at, then writes the memory range in a circular way until it gets back to where it started.

Uses separate random values for the data value to write and the address to begin at.

### DIFF
--- a/embed/boardloader/startup.s
+++ b/embed/boardloader/startup.s
@@ -18,17 +18,24 @@ reset_handler:
   bl rng_read
   mov r4, r0            // save TRNG output in r4
 
+  // r0 - value returned from previous call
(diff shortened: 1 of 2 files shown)


prusnak:
@jhoenicke What do you think? Is this something that makes sense? It complicates the code a lot and we are already quite paranoid as we erase the memory with random number instead of zeroes.

mcudev:
Closing; this is not that important, was just an idea.

PR #33, opened 2017-10-10
jump_to: reset registers to zero

mcudev:
we zero out memory, might as well get the registers too. the motivation behind this was that i wanted to make sure that the rng data used for memory clearing was not leaked to a later stage.


prusnak:
👍

PR #35, opened 2017-10-10
doc: STLINK_VER and boardloader description

mcudev:
STLINK_VER related documentation update and some notes about the boardloader


prusnak:
Thanks!

Issue #34, opened 2017-10-10
boardloader: clear peripheral local memories?


mcudev:
I'm wondering if it is worthwhile for the boardloader to clear the local memories of used peripherals when possible?

I assume that I'd have to at least enable the peripheral clocks to (and maybe the whole peripheral?) to be able to do this. I haven't tried yet. If it's worthwhile, I can spend the time to figure it out.

I did some research and found the following local memories where this might apply. I wasn't sure which USB peripheral will be used (it's changed a couple of times), so I looked up both.

**Side question**: Did you wind up using USB HS on the reference hardware? If so, did you need to add an external PHY?

 * OTG_FS:
section 34.10: "The USB system features 1.25 Kbyte of dedicated RAM with a sophisticated FIFO control mechanism."
section 34.16.1 "Direct access to data FIFO RAM"
OTG_FS base + 0x20000

 * OTG_HS:
section 35.2.1 "features a dedicated 4-Kbyte data RAM"
section 35.12.1 "Direct access to data FIFO RAM"
OTG_HS base + 0x20000

 * SDIO:
section 31.9.15: "The FIFO data occupies 32 entries of 32-bit words, from address: SDIO base + 0x080 to SDIO base + 0xFC"

 * FSMC:
section 36.1: "A Write FIFO, 2-word long"
probably too small to worry about and should be covered by the display clearing code already

Note: All section references are to the manual RM0090.

prusnak:
Answer to side question:
We use HS_IN_FS, which means we do not use an external PHY. The only reason why we use HS, not FS, is higher number of user endpoints (IRC 10 instead of 6).

I think it would be worth file if you looked into this - whether it is possible to clear the periph memory without enabling it completely. Then we can discuss further.

prusnak:
Closed via https://github.com/trezor/trezor-core/pull/36

PR #36, opened 2017-10-10
boardloader: issue #34

mcudev:
I looked into the peripherals mentioned in https://github.com/trezor/trezor-core/issues/34. Only the USB OTG HS seemed interesting as far as harboring data.
I found that I needed to enable the clock to the peripheral in order to read and write the data, but that was it. I didn't have to set it up more. So, clock it, write the local memory, and unclock it.
I don't think that leaves any unintended side-effects. It doesn't appear to as far as I tested.

Here's a good test:
 1. in gdb, find nothing with: `find 0x40060000,0x4007ffff,"arise! chikun arise!"`
 1. `./trezorctl ping "arise! chikun arise!"`
 1. in gdb, find a bunch of references with the same find command from step 1
 1. Reset the device and step through and pay attention to when the data becomes visible and when it goes away. You'll notice that overwriting the 4KB block removes all of the references due to how the FIFO memory is mapped.

### DIFF
--- a/embed/boardloader/main.c
+++ b/embed/boardloader/main.c
@@ -130,6 +130,7 @@ void check_and_jump(void)
 
 int main(void)
 {
+    clear_peripheral_local_memory();
     periph_init();
 
     if (0 != display_init()) {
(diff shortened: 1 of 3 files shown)


prusnak:
Thanks!

PR #37, opened 2017-10-10
docs: dev kit hardware updates mainly usb

mcudev:
Mainly updating the dev kit info to use USB HS. Some other minor updates too.


prusnak:
Thanks!

PR #40, opened 2017-10-12
stm32f427 vector table updates

mcudev:
First, I changed some zero definitions to `architecture_reserved_handler` which just winds up setting the `default_handler` for those entries. This is just a little cleaner to me. Technically, the zero definitions are invalid entries that would cause a UsageFault and wind up at the `default_handler` anyways because according to the ARMv7-M manual section B1.5.3, "entries
must have bit[0] set to 1, because this bit defines the EPSR.T bit on exception entry" and "On exception entry, if bit[0] of the associated vector table entry is set to 0, execution of the first instruction causes
an INVSTATE UsageFault".

I also added the new stm32f427 handler entries that all default to the default handler.

### DIFF
--- a/embed/trezorhal/vectortable.s
+++ b/embed/trezorhal/vectortable.s
@@ -14,7 +14,7 @@ default_handler:
   .endm
 
   // Reference:
-  // Table 61 - STM32F405 Reference manual (RM0090)
+  // Table 62 - STM32F427 Reference manual (RM0090)
   // Section B1.5 - ARMv7-M Architecture Reference Manual
   .section .vector_table, "a"
 vector_table:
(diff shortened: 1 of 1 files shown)


mcudev:
Also, I'm still testing the changes to f427 from yesterday. I want to put the new board on the scope to check that the clocks are still perfect. I think I'll have some minor updates in that area.

Side note: I am having some stability issues with the updated discovery board. Some times it works fine some times it doesn't. I think it's related to having the pins that we use for the display also being attached to a SDRAM chip. I might have to desolder that to be sure.

prusnak:
👍

PR #41, opened 2017-10-12
SystemInit: add more waiting just to be sure

mcudev:
just being a little more conservative. i put this on the scope and it works fine. the prior version works too on the f427, just this seems slightly safer.


prusnak:
👍

PR #42, opened 2017-10-12
RNG: Delay after an RCC peripheral clock enabling

mcudev:
I was going through the STM32F42xx and STM32F43xx Errata sheet and found this: 2.1.6 "Delay after an RCC peripheral clock enabling". The HAL handles it better. Let's use that here.

@prusnak  heads-up to check this one in the errata sheet too:
"2.1.11 Data cache might be corrupted during Flash read-while-write operation"
Not sure if that affects how you were going to use the flash (the memory layout docs that you updated yesterday came to mind when I saw this one).

### DIFF
--- a/embed/boardloader/lowlevel.c
+++ b/embed/boardloader/lowlevel.c
@@ -11,11 +11,12 @@
 
 void clear_otg_hs_memory(void)
 {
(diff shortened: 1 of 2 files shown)


mcudev:
added a second commit for `clear_otg_hs_memory` too since it could be affected the same way.
in testing, these don't seem to matter, but maybe in production it would. anyways, these changes lean conservative.

prusnak:
👍

PR #43, opened 2017-10-13
boardloader: display_init stm32f42x updates

mcudev:
Please try this for DISPLAY_ST7789V. It works much better for DISPLAY_ILI9341V.

This seems to cure the weird display and general system issues that I experienced. I think they must have been tied to an async exception being triggered (one that was killing the debugger even).
Notably, this changes the memory type from `FMC_MEMORY_TYPE_NOR` to `FMC_MEMORY_TYPE_SRAM`.

I moved the display pwm init earlier to avoid more screen flicker.

The change to the CMD=0xEA is just because it only takes one parameter according to the DISPLAY_ILI9341V manual.


prusnak:
This does not work for my DISPLAY_ST7789V display. Will look into that further tomorrow.

mcudev:
ok, thanks. it works fantastic for my DISPLAY_ILI9341V right now. hopefully it's just parameter value differences. also, i was thinking that after this works, you can use your gamma values with DISPLAY_ST7789V.

mcudev:
added a 2nd commit that removes the extra tearing effect / DISPLAY_VSYNC stuff because my display module supports that and i added the ILI9341V command that makes it work now.

prusnak:
After changing DataSetupTime from 1 to 5 (original value) the display started to work correctly.

prusnak:
I tested various combinations of possible values and these are the minimum values I am able to use: https://github.com/trezor/trezor-core/commit/b6f9df10bb2bd99db85bcd2fa91c668fea59e312

If I decrease more any of the values, glitches start to appear on display.

mcudev:
Do any of the ones that I had marked as "don't care" matter for you? They're not supposed to matter according to table 266 in RM0090.

IIRC AddressSetupTime may have mattered the lower I took it, and that's why I set it to max at 15 (to be conservative). If you can live with that higher, it may be safer. It's a little magical to me because the display driver manual/datasheets are nearly garbage.

DataSetupTime really only affected how quickly my display drew itself. At 255, it was noticeably slower than 1. That's why I chose 1. 4 is fine.

b6f9df10bb2bd99db85bcd2fa91c668fea59e312 appears to work for me (doing some quick tests).

prusnak:
BusTurnAroundDuration, CLKDivision and DataLatency really don't matter, so I set them to lowest values allowed by RM0090

First three did matter, what is weird that according to my display datasheet the values shoud be:

AddressSetupTime: 0 ns
AddressHoldTime: 10 ns
DataSetupTime: 10 ns

If I set them to 0, 2, 2 (because 1 AHB clock cycle is 5.95 ns @ 168 MHz) it did not work, unless I set the address setup time to 4 clocks, then I could decrease to AddressHoldTime (lowest value allowed by RM0090) or even 0.

Original numbers (2, 0, 5) also worked ...

mcudev:
Interesting, I'll take another look at the manuals later to try to make sense of it. I think DataSetupTime is a +1 value for writes, or something.

mcudev:
Ha, yeah the timing characteristics don't exactly add up.
For my display, it's the same as yours except it has a 0ns AddressHoldTime specified.

These were actually just working for me:

```
    normal_mode_timing.AddressSetupTime = 1;
    normal_mode_timing.AddressHoldTime = 1;
    normal_mode_timing.DataSetupTime = 1;
```

I'm guessing that you need a higher AddressSetupTime value due to the Mode1 write timing diagram not having a portion for AddressHoldTime. So that just gets handled as "AddressSetupTime".

AddressSetupTime = 0 does not work for me.

I'm guessing DataSetupTime = 1 works because it actually gets one more HCLK added on, giving it the 2 that we supposedly need.

Have you tried this?:

```
    normal_mode_timing.AddressSetupTime = 4;
    normal_mode_timing.AddressHoldTime = 1;
    normal_mode_timing.DataSetupTime = 1;
```

prusnak:
Yes, 4-1-1 does not work. 4-1-4 does.

PR #44, opened 2017-10-13
SConscript: trivial string concat cleanup

mcudev:
trivial cleanup


mcudev:
I just added a second commit that builds on the first.
it adds the freestanding flag and links against the nano c library instead of using the micropython dependency. it's a little smaller. seems to work.

prusnak:
👍

PR #46, opened 2017-10-14
startup: setup the stack protector for the stage before main function

mcudev:
setup the stack protector for the stage before main function


prusnak:
👍

PR #47, opened 2017-10-14
docs: bootloader microsd creation note and typo fix

mcudev:
add notes on how to manually create a bootloader microSD and fix a typo.

i was going to make this a make target, but then figured that i didn't want to make it a dangerous footgun tool.


prusnak:
👍

PR #48, opened 2017-10-14
code cleanup

mcudev:
It was easier to do this all together.

 * disable stack protector for the 3 rng functions: rng_init, rng_read, and rng_get
 * change return type of rng_init to void
 * in multiple spots, i added type qualifiers to unsigned literals (just being pedantic)
 * in stm32.c, the previous way of disabling the stack protector was still applying to all the functions in the file, no matter what i tried. so i switched to the explicit pragma for the whole file (reference gcc manual section 6.61.15).
 * changed the type of `cfgr` from int to uint32_t (just being pedantic)
 * removed the fpu coprocessor preprocessor stuff since it's always been enabled and always will be
 * `SysTick_Handler` removed some bad comments. the machine code generated is actually the same. we want volatile specified so that reads of the shared variable don't miss writes too. we do not want the compiler optimizing updates to uwTick so that only the handler sees them.

### DIFF
--- a/SConscript.boardloader
+++ b/SConscript.boardloader
@@ -104,7 +104,7 @@ env.Replace(
     '-fstack-protector-all -ffreestanding '
     + CCFLAGS_MOD,
     CCFLAGS_QSTR='-DNO_QSTR -DN_X64 -DN_X86 -DN_THUMB',
-    LINKFLAGS='-nostdlib -T embed/boardloader/memory.ld --gc-sections',
+    LINKFLAGS='-nostdlib -T embed/boardloader/memory.ld --gc-sections -Map=build/boardloader/boardloader.map',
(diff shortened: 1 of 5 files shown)


mcudev:
added a 2nd commit that:

* explicitly places `__stack_chk_guard` and `uwTick` in .bss by setting to zero (instead of making the linker decode to put them there -- otherwise COMMON)
* output boardloader linker map for ease of code review
* further simplified `SysTick_Handler` since `sys_tick_get_microseconds` does not exist in the code base. the code was vestigial.

mcudev:
added 3rd commit with minor systick handler updates that add a note about timer wrap and use the increment operator instead of `+=1`

prusnak:
Thank you!

PR #51, opened 2017-10-16
display: remove redundant display commands and increase wait after reset

mcudev:
These are some easy simplifications that remove redundant commands due to software reset being used.

Also, the default value of power on and software reset is sleep-in mode. So, this applies:
"If software reset is sent during sleep in mode, it will be necessary to wait 120msec before sending sleep out command."

Section 15.1 in this manual, http://www.buydisplay.com/download/ic/ILI9341.pdf, is very helpful. So are the application notes at the end of the manual.

I'll have further changes in this area but these were the first simple set and I wanted to keep them easy to review.

### DIFF
--- a/embed/extmod/modtrezorui/display-stm32.h
+++ b/embed/extmod/modtrezorui/display-stm32.h
@@ -227,8 +227,7 @@ int display_init(void) {
     HAL_Delay(5);
 #if DISPLAY_ILI9341V
     CMD(0x01);                                      // software reset
-    HAL_Delay(20);
-    CMD(0x28);                                      // display off
+    HAL_Delay(120);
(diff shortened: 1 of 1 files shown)


prusnak:
Here is the [datasheet for ST7789V](https://github.com/trezor/trezor-core/files/1387422/ST7789V.pdf). It seems it contains exactly the same sentences: "If software reset is sent during sleep in mode, it will be necessary to wait 120msec before sending sleep out command.", etc.

mcudev:
Yep, agreed. I double checked these changes in both manuals. I just don't have the other display to test. These work with my ILI9341.

prusnak:
Send me your address to stick@satoshilabs.com and I'll get you a dev board delivered to you!

Issue #49, opened 2017-10-14
boardloader: todo SDIO_IRQHandler


mcudev:
SDIO_IRQHandler is only possibly used by the firmware. It's not used by the boardloader and the bootloader.

It would be nice to not link it with the boardloader so that it does not get into the vector table.
But, need to move to a separate file and maybe disable stack protector and maybe mark `sd_handle` volatile?

mcudev:
addressed by https://github.com/trezor/trezor-core/pull/53

PR #53, opened 2017-10-16
SDIO: remove SDIO_IRQHandler and interrupt config

mcudev:
Addressing https://github.com/trezor/trezor-core/issues/49

With these changes, the boardloader vector table is clean.
I've tested that the boardloader SDIO stuff works with this code removed. I was able to put a new bootloader on and install a firmware using that (Nice work! Those are changes from today that I haven't reviewed other than to run functional tests so far).

The bootloader already did not have this interrupt handler in its vector table.

That left the firmware as the only place that might use SDIO interrupts. Upon further inspection, I don't think we use SDIO interrupts anywhere, so I've removed the IRQ setup etc...

### DIFF
--- a/embed/trezorhal/sdcard.c
+++ b/embed/trezorhal/sdcard.c
@@ -30,9 +30,6 @@
 
 #include "sdcard.h"
 
-#define IRQ_PRI_SDIO            4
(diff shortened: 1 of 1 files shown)


prusnak:
Thanks!

PR #54, opened 2017-10-16
boardloader, bootloader, firmware: only call periph_init in boardloader

mcudev:
don't need to call `periph_init` from each stage. just call it once in the boardloader.

future PR todos:
 * i still have to research the flash caching to see if that's desirable for this particular application. (errata notes)

 * some of the other things going on in periph_init also seem unnecessary.

 * might further consolidate where the clocks are enabled (right now it's split between here and `display_init` for example).

### DIFF
--- a/embed/boardloader/main.c
+++ b/embed/boardloader/main.c
@@ -1,3 +1,5 @@
+#include STM32_HAL_H
+
 #include <string.h>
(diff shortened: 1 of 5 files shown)


prusnak:
:+1:

PR #55, opened 2017-10-17
display: handle TEON and COLMOD the same for both display types

mcudev:
trivial change to make the startup sequence the same for both display types for these commands.
order does not matter to the display driver for these. just makes the code a little easier to review being in numerical order. i did test these on my ILI9341.

i'm still reviewing the reset of the display commands. the manuals must have errors because they show examples using parameter values that are invalid according to the command + parameter specifications. still trying to make sense of it...


prusnak:
:+1:

PR #56, opened 2017-10-17
Check PIN in C

jpochyla:
Not ready for merge yet, mostly PoC of T1 approach.

- [x] Pull norcow to trezor-core tree.
- [x] Port norcow on top of new flash API, remove unix-specific code.
- [x] Move PIN-related code out of modtrezorconfig.
- [ ] Figure out a better way of storing PIN (outside of norcow).
- [ ] Review everything.
- [ ] Unit test everything.


embed/extmod/modtrezorio/modtrezorio-flash.h
```diff
@@ -5,13 +5,7 @@
...
  */
 
-#if defined TREZOR_STM32
-#include "flash.h"
-#elif defined TREZOR_UNIX
-#include "unix-flash-mock.h"
-#else
-#error Unsupported TREZOR port. Only STM32 and UNIX ports are supported.
-#endif
+#include "../../trezorhal/flash.h"
```
prusnak: Do we need relative path here?

jpochyla: trezorhal is not in include path for the unix port, but we can add it.

prusnak: Ah, right. Let's keep it that way.

prusnak:
Merged in 5ffea33c3dffd95806a48827aafa0052f4860f51 :tada:

PR #59, opened 2017-10-18
bootloader, firmware: do not need touch interrupt handlers or to enable touch interrupts

mcudev:
Removing these do not change any functionality for me. These changes do rid the bootloader and firmware vector tables of the entries. I'll probably remove the USB HS interrupt handlers too, but I need to test that.

Generally, touch does not work well for me.
The bootloader touch click to continue is reliable.
But, the bootloader press to stay in bootloader mode (https://github.com/trezor/trezor-core/blob/master/embed/bootloader/main.c#L262-L268) does not work at all for me (both before and after this change).

### DIFF
--- a/embed/trezorhal/touch.c
+++ b/embed/trezorhal/touch.c
@@ -14,8 +14,8 @@ I2C_HandleTypeDef i2c_handle = {
     .Instance = I2C1,
 };
 
-int touch_init(void) {
-
+int touch_init(void)
+{
     // Enable I2C clock
     __HAL_RCC_I2C1_CLK_ENABLE();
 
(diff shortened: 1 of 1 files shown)


prusnak:
Incidentally, I am looking into the same issue. As you could've guessed from my [commit](https://github.com/trezor/trezor-core/commit/c689b1f2cd4217479a1103bc8e103ec542f28997) the touch does not work for me, neither.

I will merge this and continue investigating the issue.

mcudev:
OK, I'll look into it some too. I'm using a FT6206 capacitive touch panel controller. Is that the same as yours?

prusnak:
We are using FT6236, but I would expect they are very similar if not the same.

prusnak:
Ah, it was a stupid mistake, I'll fix it.

prusnak:
Fixed in 749b80ba030e8a3debc2612a26b7378c33b642c3

mcudev:
yes, definitely better. i have to do more than statically touch. i have to do a move also and it has missed it a few times. but the majority of the time it works to get me into bootloader mode.

mcudev:
off-topic but kind of related because of the bootloader: I had to do `sudo update-usbids` to get more than just the `InterBiometrics` info in my `lsusb` output. Now it shows: `1209:53c0 InterBiometrics SatoshiLabs TREZOR Bootloader`.
I'm also wondering if you have requested to have the USB ID list in chromium updated (they are way behind the source)? https://chromium.googlesource.com/chromium/src/+/lkgr/third_party/usb_ids

prusnak:
No, have not contacted Google.

PR #61, opened 2017-10-19
 image_parse_header: remove undefined behavior and other updates

mcudev:
first commit simply removes the undefined behavior.

second commit, updates the function declaration and changes to not use memcpy as much.

switched one comparison to `& 511` instead of `% 512`

added a todo to do more with the expiry field. maybe what it is compared to should be an argument to the func too?

### DIFF
--- a/embed/trezorhal/image.c
+++ b/embed/trezorhal/image.c
@@ -29,13 +29,8 @@ static bool compute_pubkey(uint8_t sig_m, uint8_t sig_n, const uint8_t * const *
     return 0 == ed25519_cosi_combine_publickeys(res, keys, sig_m);
 }
 
-bool image_parse_header(const uint8_t *data, uint32_t magic, uint32_t maxsize, image_header *hdr)
+bool image_parse_header(const uint8_t * const data, const uint32_t magic, const uint32_t maxsize, image_header * const hdr)
 {
-    if (!hdr) {
-        image_header h;
-        hdr = &h;
-    }
-
     memcpy(&hdr->magic, data, 4);
     if (hdr->magic != magic) return false;
 
--- a/embed/trezorhal/image.c
+++ b/embed/trezorhal/image.c
@@ -38,11 +38,13 @@ bool image_parse_header(const uint8_t * const data, const uint32_t magic, const 
     if (hdr->hdrlen != HEADER_SIZE) return false;
 
     memcpy(&hdr->expiry, data + 8, 4);
+    // TODO: expiry mechanism needs to be ironed out before production or those
+    // devices won't accept expiring bootloaders (due to boardloader write protection).
     if (hdr->expiry != 0) return false;
 
     memcpy(&hdr->codelen, data + 12, 4);
-    if (hdr->hdrlen + hdr->codelen < 4 * 1024) return false;
-    if (hdr->hdrlen + hdr->codelen > maxsize) return false;
+    if (hdr->codelen > (maxsize - hdr->hdrlen)) return false;
+    if ((hdr->hdrlen + hdr->codelen) < 4 * 1024) return false;
     if ((hdr->hdrlen + hdr->codelen) % 512 != 0) return false;
 
     memcpy(&hdr->version, data + 16, 4);
--- a/embed/trezorhal/image.c
+++ b/embed/trezorhal/image.c
@@ -76,17 +76,13 @@ bool image_check_signature(const uint8_t *data, const image_header *hdr, uint8_t
     return 0 == ed25519_sign_open(hash, BLAKE2S_DIGEST_LENGTH, pub, *(const ed25519_signature *)hdr->sig);
 }
 
-bool vendor_parse_header(const uint8_t *data, vendor_header *vhdr)
+bool vendor_parse_header(const uint8_t * const data, vendor_header * const vhdr)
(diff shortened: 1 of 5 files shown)


embed/trezorhal/image.c
```diff
@@ -29,27 +29,29 @@ static bool compute_pubkey(uint8_t sig_m, uint8_t sig_n, const uint8_t * const *
...
+    if (hdr->expiry != 0) return false; // todo: then what is the point of having this field?
 
-    memcpy(&hdr->codelen, data + 12, 4);
-    if (hdr->hdrlen + hdr->codelen < 4 * 1024) return false;
-    if (hdr->hdrlen + hdr->codelen > maxsize) return false;
-    if ((hdr->hdrlen + hdr->codelen) % 512 != 0) return false;
+    hdr->codelen = *data_ptr++;
+    if ((hdr->hdrlen + hdr->codelen) < 4 * 1024) return false;
+    if ((hdr->hdrlen + hdr->codelen) > maxsize) return false;
+    if (((hdr->hdrlen + hdr->codelen) & 511) != 0) return false;
```
prusnak: I don't like this, it is much more cryptic and the original intention is more hidden. I was under impression that `% x` will get optimized to `& (x-1)` in case x is 2^n. If that's not the case, leave it, but in case it gets optimized, I strongly prefer keeping the original way.

embed/trezorhal/image.c
```diff
@@ -29,27 +29,29 @@ static bool compute_pubkey(uint8_t sig_m, uint8_t sig_n, const uint8_t * const *
     return 0 == ed25519_cosi_combine_publickeys(res, keys, sig_m);
 }
 
-bool image_parse_header(const uint8_t *data, uint32_t magic, uint32_t maxsize, image_header *hdr)
+bool image_parse_header(const uint8_t * const data, const uint32_t magic, const uint32_t maxsize, image_header * const hdr)
 {
-    memcpy(&hdr->magic, data, 4);
+    uint32_t *data_ptr = (uint32_t *) data;
```
prusnak: Should not this be const as well?

embed/trezorhal/image.c
```diff
@@ -29,27 +29,29 @@ static bool compute_pubkey(uint8_t sig_m, uint8_t sig_n, const uint8_t * const *
...
+    if ((hdr->hdrlen + hdr->codelen) > maxsize) return false;
+    if (((hdr->hdrlen + hdr->codelen) & 511) != 0) return false;
 
-    memcpy(&hdr->version, data + 16, 4);
+    hdr->version = *data_ptr++;
 
     // uint8_t reserved[427];
 
-    memcpy(&hdr->sigmask, data + 0x01BF, 1);
+    hdr->sigmask = *(data + 0x01BF);
```
prusnak: I was using memcpy instead of assignement to avoid unaligned access, which we had problem with in the past. Are you sure this will not cause some issues? (The question is valid for all memcpy called replaced with assignments, not just this one).

embed/trezorhal/image.c
mcudev: ok, will revert. same code is generated. this was just a habit of mine to avoid the remainder operator. it does not matter here.

embed/trezorhal/image.c
mcudev: yep

embed/trezorhal/image.c
mcudev: interesting concern. i think the pointer is ok because the data pointer is aligned in current uses, and the cortex-m4 processor supports unaligned access for the LDR instruction (i tried an unaligned access and it did not blow up). but to be safe, i'll revert these changes back to memcpy to be extra safe.

mcudev:
Updates made. I also added some updates for the image sizes. I wasn't sure whether to add the size for the 7 "firmware extra" sectors (to make it 13 instead of 6 sectors), but decided not to right now to be safe because of how `_read_payload` is currently implemented (doesn't look like it would skip the sector gap to the extra ones correctly).

embed/trezorhal/image.c
```diff
@@ -38,11 +38,13 @@ bool image_parse_header(const uint8_t * const data, const uint32_t magic, const
...
 
     memcpy(&hdr->expiry, data + 8, 4);
+    // TODO: expiry mechanism needs to be ironed out before production or those
+    // devices won't accept expiring bootloaders (due to boardloader write protection).
     if (hdr->expiry != 0) return false;
 
     memcpy(&hdr->codelen, data + 12, 4);
-    if (hdr->hdrlen + hdr->codelen < 4 * 1024) return false;
-    if (hdr->hdrlen + hdr->codelen > maxsize) return false;
+    if (hdr->codelen > (maxsize - hdr->hdrlen)) return false;
```
prusnak: Is there a reason why not keep `((hdr->hdrlen + hdr->codelen) > maxsize)` here to match the following conditions?

mcudev: I was thinking about large values of codelen that could overflow when added to hdrlen and this way seemed a little more intuitive to me, that's all. As long as that first (old) comparison to 4096 uses >= 512, you're bounded/ok. The 4096 seemed a little magical other than that.

prusnak:
Thanks!

PR #62, opened 2017-10-19
vendor_parse_header: remove undefined behavior add todos

mcudev:
title says it all.
the undefined behavior is when using a pointer to a variable that has gone out of scope.

### DIFF
--- a/embed/bootloader/main.c
+++ b/embed/bootloader/main.c
@@ -266,14 +266,14 @@ int main(void)
         hal_delay(1);
     }
 
+    vendor_header vhdr;
+
     // start the bootloader if user touched the screen or no firmware installed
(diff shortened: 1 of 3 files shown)


prusnak:
Can you please add this commit into pull request #61?

Few comments: 

1) please change the prototype to ` bool vendor_parse_header(const uint8_t * const data, vendor_header * const vhdr)` so it matches the `image_parse_header` prototype

2) remove " // todo: then what is the point of having this field?" - the point is when there is a safe way to get current time into trezor (for example as a signed message from NTP server) we could check for expiration of firmware. For now just accept firmwares that don't expire.

3) use capitalization of `todo` for sanity check comment, that's the form what most tools are using

PR #64, opened 2017-10-19
boardloader, flash: flash option bytes updates, remove flash_init

mcudev:
First, I trivially removed the empty `flash_init` function from the boardloader, bootloader, and firmware.

Second, I updated the flash option byte code. Notice that I changed the `opts.OptionType =` to `opts.OptionType |=`. After looking through the HAL code, I believe the new way is the correct usage.
I also made the option byte setting code erase other non-boardloader flash sectors, attempt to write the option bytes, and then loop again to check, repeating until they are set. Untested code, but probably close.

### DIFF
--- a/embed/boardloader/lowlevel.c
+++ b/embed/boardloader/lowlevel.c
@@ -1,5 +1,6 @@
 #include STM32_HAL_H
 
+#include "flash.h"
 #include "lowlevel.h"
 
 #define WANTED_WRP (OB_WRP_SECTOR_0 | OB_WRP_SECTOR_1 | OB_WRP_SECTOR_2)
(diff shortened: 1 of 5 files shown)


prusnak:
I removed the sector erasing from flash_set_option_bytes for now. While it is generally a good idea, there is a disadvantage it makes it impossible to flash {boardloader + bootloader + testing firmware} on brand new board during the production, because on first run the bootloader + testing firmware would be deleted. Also we check whether the option bytes were set correctly afterwards via the `flash_check_option_bytes` call and then we call `flash_erase_sectors` if something is wrong.

mcudev:
OK, thanks. I was mainly focused on giving multiple opportunities to clear the storage sectors in that situation. The other sectors were "might as well, while I'm doing this already". The only times when this should apply are at production time, and if there's an attack like the one described in "Shedding too much Light on a Microcontroller’s Firmware Protection".
Are you going to ship v2 without firmware like v1? If so, does the testing firmware make changes to the firmware sectors before it completes? The testing firmware could run from SRAM.

prusnak:
Yes, v2 will ship with boardloader and bootloader, no firmware.

Testing firmware does not write to flash (except OTP area, where date of production is stored).

PR #65, opened 2017-10-22
boardloader: enable PVD programmable voltage detector and interrupt, don't clear standby flag

mcudev:
The peripherals used, especially, but not limited to, the display, preclude level 6 and 7 on my dev board. The risk of enabling this detector, and continuing to use the BOR level 3 option byte, is that they affect regular users (maybe longer USB cables, bad USB ports, etc... causing voltages closer to the thresholds). The reason to enable these thresholds is to provide a level of defense against voltage glitching attacks, and/or poorly functioning components (similar to the CSS does for the HSE clock).

Level 5 seems to work for me. I tested the regular startup and also a boardloader microSD reflash.
Level 7 trips in the boardloader for me while wiping the USB HS memory. Level 6 trips in the firmware while drawing the initial screen (if it doesn't then try an all white trezor logo with full backlight and see if it does).

I left the `PVD_IRQHandler` mapped to `default_handler` as the intent of this is to hinder voltage glitching attacks.

Side note: it maybe worthwhile profiling the power supply on actual devices to see how they hold up when full current draw is made in unison by the display, flash, cpu, etc..

### DIFF
--- a/embed/boardloader/lowlevel.c
+++ b/embed/boardloader/lowlevel.c
@@ -76,7 +76,20 @@ void periph_init(void)
     HAL_PWR_ConfigPVD(&pvd_config);
     HAL_PWR_EnablePVD();
     NVIC_EnableIRQ(PVD_IRQn);
+}
 
-    // Clear the reset flags
-    RCC->CSR |= RCC_CSR_RMVF;
+bool reset_flags_init(void)
+{
+#if PRODUCTION
+    // this is effective enough that it makes development painful, so only use it for production.
+    // check the reset flags to assure that we arrive here due to a regular full power-on event,
+    // and not as a result of a lesser reset.
+    if ((RCC->CSR & (RCC_CSR_LPWRRSTF | RCC_CSR_WWDGRSTF | RCC_CSR_IWDGRSTF | RCC_CSR_SFTRSTF | RCC_CSR_PORRSTF | RCC_CSR_PINRSTF | RCC_CSR_BORRSTF)) != (RCC_CSR_PORRSTF | RCC_CSR_PINRSTF | RCC_CSR_BORRSTF)) {
+        return false;
+    }
+#endif
+
+    RCC->CSR |= RCC_CSR_RMVF; // clear the reset flags
+
+    return true;
 }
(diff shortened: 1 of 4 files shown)


mcudev:
Second commit: there is no reason to enter standby mode (we don't that i know of), therefore there is no reason to mask the fact that we were in standby mode (if somehow that happened)

mcudev:
Third commit: make use of the reset flags to check for unusual conditions and bail out if encountered. also updated the boardloader startup code to branch to `shutdown`.

helpful for testing this (in the `Makefile`):
`OPENOCD = openocd -f interface/stlink-$(STLINK_VER).cfg -c "transport select hla_swd" -f target/stm32f4x.cfg -c "reset_config srst_only separate srst_nogate srst_open_drain connect_assert_srst"`

mcudev:
Fourth commit: do not use ensure before display_init, just return (since the previous commit returning calls `shutdown`).

mcudev:
Fifth commit: another avoid calling `shutdown` directly, and let the startup code handle it

mcudev:
@prusnak Just checking to see if you did not like these, or if you're just too busy to review right now?

prusnak:
I will review soon. Too busy last 2 days.

prusnak:
Very nice. I have only one comment to https://github.com/trezor/trezor-core/pull/65/commits/23f86b5c6ec236207c957da27d55bfc2d0b8df4f

We might want to allow switching from bootloader to firmware (after successful firmware update) and from firmware to bootloader (so one can "reboot" to bootloader from the sofware UI).

I THINK this might not break this functionality, but I am not really sure. Can you confirm that this change will not break this intention?

mcudev:
It depends how that stage switching is implemented. If it's via branches/jumps, then likely no problem. If it's via something like `NVIC_SystemReset`, then that reset flag checking code would block it in production. The bootloader already continues on to new firmware on successful update, so it's really just a question of how a firmware back to bootloader would work.

BTW, in quick tests of bootloader to firmware switch after firmware update, I think that there is a bug with USB preventing that from working all the way. I can see that the firmware takes over, but that it gets stuck in the USB code. Probably the bootloader endpoint still being active causing that, but I have not looked closely enough.

prusnak:
Correct, I plan to use `jump_to`, not SystemReset.

Correct two that the current jump from bootloader to firmware does not work atm. :-/

mcudev:
should be fine then.

it seems a little dangerous to do the code updates and then switch immediately, but i can't think of a specific vulnerability right now (maybe sectors don't erase? but now with the extra voltage detection, that's even less likely). definitely an area worthy of extra review when it's ready.

PR #68, opened 2017-10-24
bootloader: cleanup usb on firmware upload success before jumping to firmware

mcudev:
This fixes the firmware hanging after successful upload by the bootloader.

Not sure if this is the best way to handle this, but it seems to work. Notably, this works best after the 3 second wait. Otherwise, IO read errors are encountered by `trezorctl`.

Tested with `lsusb` and `python-trezor> ./trezorctl firmware_update -f ../trezor-core/build/firmware/firmware.bin`

### DIFF
--- a/embed/bootloader/main.c
+++ b/embed/bootloader/main.c
@@ -226,6 +226,8 @@ bool bootloader_loop(void)
                     display_done(3); hal_delay(1000);
                     display_done(2); hal_delay(1000);
                     display_done(1); hal_delay(1000);
+                    usb_stop();
+                    usb_deinit();
                     display_fade(BACKLIGHT_NORMAL, 0, 500);
                     return true; // jump to firmware
                 }


prusnak:
:+1:

prusnak:
Added the same deinit+stop combo to other options where device is supposed to be unplugged (after successful wipe or failed wipe/upgrade) in 9e31e0f4d4175bceaf7decc2a7c537c8414095d0.

I thought it would help with USB IO errors after the device has called shut down, but they are still there. Oddly enough, I see IO errors even if I call `hal_delay(3000);` before usb_stop+usb_deinit combo or after the combo. :-/

mcudev:
ok, i'll keep that in mind. the usb code is a weak area for me that i need to review more to understand.

prusnak:
Added calls to USBD_LL_FlushEP in 8bd258db6f2d5ac0395629013eac9e5591febea9 but it does not help anyway. I guess we need to wait with WFI or something like that.

mcudev:
> I guess we need to wait with WFI or something like that.

That could be. I only reviewed the USB code as far as to know that the `OTG_HS_IRQHandler` is used and when I was doing that, I remember seeing code doing exactly that.

PR #72, opened 2017-10-27
display: driver cleanup

mcudev:
I'd like for someone to try this with ST7789V and let me know if it fails.
I don't think it'll be too far off since it works great for ILI9341.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,9 @@ flash_erase: ## erase all sectors in flash bank 0
 openocd: ## start openocd which connects to the device
 	$(OPENOCD)
 
+openocd_reset: ## cause a system reset using OpenOCD
+	$(OPENOCD) -c "init; reset; exit"
+
 GDB = arm-none-eabi-gdb --nx -ex 'set remotetimeout unlimited' -ex 'set confirm off' -ex 'target remote 127.0.0.1:3333' -ex 'monitor reset halt'
 
 gdb_boardloader: $(BOARDLOADER_BUILD_DIR)/boardloader.elf ## start remote gdb session to openocd with boardloader symbols
--- a/embed/bootloader/main.c
+++ b/embed/bootloader/main.c
@@ -290,8 +290,6 @@ int main(void)
     check_bootloader_version();
 #endif
 
-    display_orientation(0);
-
     touch_init();
 
     // delay to detect touch
--- a/embed/extmod/modtrezorui/display-stm32.h
+++ b/embed/extmod/modtrezorui/display-stm32.h
@@ -97,7 +97,10 @@ void display_set_backlight(int val)
 void display_hardware_reset(void)
 {
     HAL_GPIO_WritePin(GPIOC, GPIO_PIN_14, GPIO_PIN_RESET); // LCD_RST/PC14
-    HAL_Delay(1); // wait 1 millisecond. only needs to be low for 10 microseconds.
+    // wait 10 milliseconds. only needs to be low for 10 microseconds.
+    // my dev display module ties display reset and touch panel reset together.
+    // keeping this low for max(display_reset_time, ctpm_reset_time) aids development and does not hurt.
+    HAL_Delay(10);
     HAL_GPIO_WritePin(GPIOC, GPIO_PIN_14, GPIO_PIN_SET); // LCD_RST/PC14
     HAL_Delay(120); // max wait time for hardware reset is 120 milliseconds (experienced display flakiness using only 5ms wait before sending commands)
 }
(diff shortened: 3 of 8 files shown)


prusnak:
The patch works fine except one detail: the picture is moved 80px to the top (in all directions):

ea26af0d-b0a0-4471-90c6-0aecd41fb592

mcudev:
Thanks! I'll fix this when I have a ST7789V to test against. I have a feeling there are a few other things that I can do to simplify at the same time.

mcudev:
@prusnak ok, these changes work now. i re-arranged the commits to try to make them easier to follow.

you might find `make openocd_reset` or `make build_boardloader flash_boardloader openocd_reset` useful, especially after flashing code, i know i do :)

this fixes some subtle bugs, for example, when orientation is such that the top of the lock is pointing right on the homescreen, running `./trezorctl wipe_device` has a incompletely drawn screen.

I'd really like to remove the ST7789V config for PWCTRL1 in `display_init`. I left it because you had it, but I don't see it doing anything that I can easily observe. Also, you can change the gamma values if you'd like. They do make noticeable changes to the appearance of the colors.

mcudev:
added a couple more commits for display and ctpm reset and usage updates

mcudev:
@prusnak rebased just now. do i need to do anything else to get this merged?

prusnak:
No, I will get this merged soon. Sorry.

mcudev:
No worries, this code is not easy to review and I didn't package it well. Also, just pointing out that the make target `openocd_reset` will probably make your testing life easier :)

prusnak:
The code works as expected, with one minor detail - the vertical orientation is inverted. (Device starts with upside down image).

mcudev:
Hmm, works for me on the stuff you sent, try this:
`make clean vendor vendorheader build_boardloader build_bootloader build_firmware flash openocd_reset`

mcudev:
Also, I was wondering if you added the "PWCTRL1: Power Control 1" line for ST7789V for a particular reason, or if you were told to put that in by the manufacturer, etc...?

prusnak:
Hah, I have two boards that produce different results (one is ok, another is upside-down). I guess we need to keep set the orientation explicitly, because it's not the same by default everywhere.

prusnak:
> if you were told to put that in by the manufacturer, etc...?

I think I copied the sequence from some example I found online. It was not from the manufacturer.

mcudev:
> two boards that produce different results

lolwut, `display_clear` still sets the orientation at the beginning. is the logo upside down in the bootloader, or just later in the firmware?

prusnak:
Both

mcudev:
and what happens with the sd card inserted (code stays in the boardloader)?
that would be weird if the screen orientation is different within a batch for a prod run.

prusnak:
Just confirmed that the board works correctly with the code in `master`.

mcudev:
The only time I can remember the logo being upside down was when I forgot to flash the subsequent stages. That's why I was curious about the SD card test (since it stays in the boardloader and thus would be the latest flash definitely).

prusnak:
Bootloader is also upside down on that particular board.

The other board is fine. 

*puzzled*

mcudev:
bootloader and firmware, if they don't get flashed with this new code, but the boardloader does, i can understand.
boardloader should display the text correctly though all the time because you'd need to flash it to get these changes. hence the sd card test.

prusnak:
> Bootloader is also upside down on that particular board.

I meant boardloader - the SD card test.

mcudev:
now that is puzzling. that sounds like a difference in the device manufacture. when holding the device upright, normally, (0,0) of the display is in the bottom right corner. the CTPM has (0,0) in the top left.

mcudev:
on the board that is upside down, please try this on the `GATECTRL` line:
`CMD(0xE4); DATA(0x1d); DATA(0x0A); DATA(0x10);`
that flips the gate scan to be the normal 0->319 mode.

prusnak:
It is still upside down in boardloader.

mcudev:
wow. does `make flash_erase` work for that board?

mcudev:
if you swap display between boards does anything change?

mcudev:
Hmm, well thanks for testing. I did not foresee this problem. I'll assume the old code works on both and my code does not and I'll try to reconcile that later.

prusnak:
`make flash_erase` funguje

prusnak:
I confirm that if I swap displays between two board, the odd behaviour is appearing on the other board.

mcudev:
Please try:
`CMD(0xC0); DATA(0x00); // LCMCTRL: LCM Control: toggle default XOR RGB setting`

If the display changes, we can rule out the hypothesis below.
If nothing changes, then the below hypothesis is worth further analysis.

Since the 0xE4 command did not have the effect that I was expecting, I have a hypothesis that the EXTC pin, which allows access to the extension commands(0xE4 is one such command), might not be high. If so, that could indicate a display module quality issue. To test this, we can try the 0xC0 command (also an extension command) and see if changing it has an effect on the display.

prusnak:
Nothing changed

prusnak:
Also I confirm that if I switched the display on the same board, the other display showed blue triangle instead of yellow one in the bootloader.

mcudev:
Exactly. Now, this is low probability, but if you add this next line just after the 0x3A COLMOD line and before the 0xC0 line, does the upside down display work?
`CMD(0xDF); DATA(0x5A); DATA(0x69); DATA(0x02); DATA(0x01);`

0xDF is an extension command, but its description says that when enabled, "Commands in command table 2 can be executed when EXTC level is “Low”."

prusnak:
Yeah! This line fixed the issue!

mcudev:
Sweet, that is an important find! Add that line and we should be good.Thanks!!!!

prusnak:
Thank you very much! You are truly a wizard! :D

mcudev:
haha, i got lucky, but i'll take it :)

PR #73, opened 2017-10-30
clear_otg_hs_memory: remove unpredictable write

mcudev:
linking in discussion https://github.com/trezor/trezor-core/commit/6738796d8b0c5eb73a3467e66483b31d4c0b4231


mcudev:
added a second commit that reorders the operations done during `shutdown`. changed to call `clear_otg_hs_memory` last just in-case there is a problem -- do everything more important first. i also considered setting everything to 0xaaaaaaaa instead of 0. but didn't feel strongly about it. if it makes sense to do that, it's a simple change.

PR #75, opened 2017-10-31
boardloader: production changes, flash option bytes, pvd level 5

mcudev:
Addressing a few things with this PR:

1. https://github.com/trezor/trezor-core/commit/311a28242b01a5617c0d4a940d6ba34ac55a2059
1. https://github.com/trezor/trezor-core/commit/16f657361c2565b6206816e0a456042f352f946b
1. https://github.com/trezor/trezor-core/commit/a08f18c2ce5b9e6518bf5fd7a989e0862d779975

I got the HAL version working, but noticed these things while doing it that made me think that what I'm proposing here is better:
1. HAL code does not actually check flash memory values
1. HAL flash code does not perform all recommended waits
1. HAL code depends on systick timer
1. HAL does not reset flash status flags

These changes also makes sure that the option byte code gets tested more as it is in the normal flow, and only a couple data values are different. I tested that setting the BOR, RDP (levels0 and 1), and WRP option bytes works.

There are a couple of subtle bug fixes mixed in.

I changed the header guards to not use reserved names.

There is a status flag addition for the 427 chip in `flash_unlock`.

Useful GDB debugging/testing commands for getting back to RDP level 0 from level 1:

```
// check FLASH->SR register
x/1xw 0x40023c0c

// check FLASH->OPTCR register
x/1xw 0x40023c14

// check FLASH->OPTCR1 register
x/1xw 0x40023c18

// unlock FLASH->OPTKEYR register
set * (unsigned int *) 0x40023c08 = 0x08192A3B
set * (unsigned int *) 0x40023c08 = 0x4C5D6E7F

// set defaults in FLASH->OPTCR1 register
set * (unsigned int *) 0x40023c18 = 0x0fff0000

// set defaults in FLASH->OPTCR register and commit changes to flash
set * (unsigned int *) 0x40023c14 = 0x0fffaaee

// check OPTION_BYTES_RDP_USER in flash memory
x/1xh 0x1fffc000

// check OPTION_BYTES_BANK1_WRP in flash memory
x/1xh 0x1fffc008

// check OPTION_BYTES_BANK2_WRP in flash memory
x/1xh 0x1ffec008
```

### DIFF
--- a/embed/boardloader/lowlevel.c
+++ b/embed/boardloader/lowlevel.c
@@ -1,53 +1,107 @@
 #include STM32_HAL_H
 
+#include "flash.h"
 #include "lowlevel.h"
 
-#define WANTED_WRP (OB_WRP_SECTOR_0 | OB_WRP_SECTOR_1 | OB_WRP_SECTOR_2)
-#define WANTED_RDP (OB_RDP_LEVEL_2)
+#pragma GCC optimize("no-stack-protector") // applies to all functions in this file
 
-void flash_set_option_bytes(void)
+#if PRODUCTION
+    #define WANT_RDP_LEVEL   (OB_RDP_LEVEL_2)
+    #define WANT_WRP_SECTORS (OB_WRP_SECTOR_0 | OB_WRP_SECTOR_1 | OB_WRP_SECTOR_2)
+#else
+    #define WANT_RDP_LEVEL   (OB_RDP_LEVEL_0)
+    #define WANT_WRP_SECTORS (0)
+#endif
+
+#define WANT_BOR_LEVEL (OB_BOR_LEVEL3)
+
+// reference RM0090 section 3.9.10; SPRMOD is 0 meaning PCROP disabled.; DB1M is 0 because we use 2MB dual-bank; BFB2 is 0 allowing boot from flash;
+#define FLASH_OPTCR_VALUE ( (((~WANT_WRP_SECTORS) << FLASH_OPTCR_nWRP_Pos) & FLASH_OPTCR_nWRP_Msk) | \
+                            (WANT_RDP_LEVEL << FLASH_OPTCR_RDP_Pos) | FLASH_OPTCR_nRST_STDBY | FLASH_OPTCR_nRST_STOP | FLASH_OPTCR_WDG_SW | WANT_BOR_LEVEL )
+
+// reference RM0090 section 3.7.1 table 16
+#define OPTION_BYTES_RDP_USER_VALUE  ((uint16_t) ((WANT_RDP_LEVEL << FLASH_OPTCR_RDP_Pos) | FLASH_OPTCR_nRST_STDBY | FLASH_OPTCR_nRST_STOP | FLASH_OPTCR_WDG_SW | WANT_BOR_LEVEL))
+#define OPTION_BYTES_BANK1_WRP_VALUE ((uint16_t) ((~WANT_WRP_SECTORS) & 0xFFFU))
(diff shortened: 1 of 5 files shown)


prusnak:
Instering SD card with BOR level 3 restarts the board.

Instering SD card with PVD level 5 halts the board.

If I set BOR to level 2 and PVD to level 4 it works without problems.

I am going to merge this, but change the values to the ones above. :-/

I am not sure if there is something we can do here, unfortunately.

PS: I am using 1 meter long USB-C cable connected directly to my laptop.

prusnak:
First commit merged in https://github.com/trezor/trezor-core/commit/1737f17f43c936041a1355baa98dd3fec8127a1c

Second commit ignored (as PVD level stays at 4).

BOR level set to 2 in https://github.com/trezor/trezor-core/commit/8190e0909e17d737d5e5cea8686f7199adf9b1df

mcudev:
Thanks for merging what you did!

I'll do more testing and scope some different code paths. It seems fairly obvious that some peripherals are drawing more current and thus causing a voltage drop due to the increased resistance of a longer USB cable.

I do feel somewhat strongly that BOR 3 and PVD 5 should be used even if it means that longer cables are not usable. At lower voltages, the device is operating functionally out of spec according to the manuals. It's better for security too if the proper voltages are enforced.

mcudev:
@prusnak I noticed that the act of inserting the microSD shorts the power supply to GND. That was unexpected. You don't even have to pushit all the way in, just play with the spring a little and BANG!.

prusnak:
> Happens on my dev kit board too even when only the 3v3 and GND are connected to microSD.

You mean the Discovery board or the board I've sent you?

prusnak:
Thanks for the investigation! I'll discuss with the team if it is acceptable to reboot on SD card insertion (could be seen as a feature).

mcudev:
You're welcome! I don't think it's a reliable feature though. Sometimes when I did it, the bootloader hanged before fading away etc... I'm not sure if the processor went into lockup, or if it was just because of oscillation in the power supply. What I fear is that that it's just a symptom pointing at the power supply needing some work though. More capacitance at least. Maybe some more advanced stuff possibly.

mcudev:
BTW, if you wind up making power subsystem changes, here are some thoughts:
* Except for power-on and power-off, you always want to be in the **range 2.7V - 3.6V; targeting 3.3V** (ref RM0090 section 3.5, 3.6.2 and datasheet tables)
* It would be nice to be able to use PVD level 7 (2.9V)
* Add MCU controlled switches to control power supply to display, touch panel, and microSD (this will allow for example, to control screen flicker at startup)
* Add any MCU recommended capacitors
* Add decoupling capacitors near microSD and maybe display

Adding two informational links:
* http://www.ti.com/lit/an/slva670a/slva670a.pdf
* Guy mentiones microSD in-rush current around 200mA: http://www.smallridge.com.au/download/SD-MMC%20Integration.pdf

mcudev:
@prusnak just following-up to see if you all have decided on any power subsystem changes?

prusnak:
For first batch of devices, it is already too late. We'll revisit the hardware for next batch.

mcudev:
dang, too bad. i did some testing with my dev kit and was able to keep the sd card from power faulting (due to in-rush current) by adding capacitors. i have more research too as i went partially through the process of creating some contingency code to try and persevere even with low voltage. it gets ugly though. maybe i should just open an issue to centralize the tracking for power related issues?

prusnak:
Yes, please do. I have a tracker issue with your comments in my internal
todo list, but having it on Github would be most probably much better.

PR #76, opened 2017-11-05
cleanup and jump_to update

mcudev:
4 self-contained commits here. The 4th is most important:

1. Updated the secbool header guards to not use reserved identifiers

1. RNG while loop codestyle, RNG header guards, and minor cleanup

1. Change to follow the typical pattern of including the CMSIS device header instead of the HAL header because the CMSIS device header will include the HAL header when USE_HAL_DRIVER is defined.

1. The last commit clears the registers before even the next stage's NMI handler can run. don't want to leak to that either. also call `clear_otg_hs_memory` before jumping. the other changes are just re-arranging / cleanup.

Side note: There are newer versions of the CMSIS and device specific files available. Do you want to use those, or stick to what comes from micropython?:

 * https://github.com/ARM-software/CMSIS_5

 * http://www.st.com/en/embedded-software/stm32cubef4.html

I almost changed "STM32_HAL_H" to "STM32_DEVICE_H" except micropython example ports use this name and norcow would need a change merged too.


prusnak:
Let's stick to the ones used by uPy if there is nothing really interesting in newer versions. (ATM looking into CMSIS 5, but haven't found anything worth yet ...)

PR #78, opened 2017-11-06
boardloader: ensure check_sdcard fails when read fails

mcudev:
During testing, I was curious why disabling `INSTRUCTION_CACHE_ENABLE` caused the SD card code to fail. It turns out that it causes SDIO Rx FIFO overruns (my guess is that it is due to the extra flash waits introduced by instruction loads from flash). At least it does not look to be caused by SD power draw.

These code changes ensure that the checks fail when the read fails.


mcudev:
crap, don't take this yet, i malfunctioned

mcudev:
ok, this is ready now

prusnak:
Nice catch!

Issue #77, opened 2017-11-06
Memory allocation failed


dominik-selmeci:
Hi, 
I'm learning how to work with the emulator. I used [this manual](https://github.com/trezor/trezor-core/blob/master/docs/build.md#debianubuntu). The build was without errors. When I access emu.sh through terminal id outputs:

> Traceback (most recent call last):
>   File "main.py", line 112, in <module>
> MemoryError: memory allocation failed, allocating 136 bytes

My system:

> Kubuntu 17.10 64bit (plasma)
> Python 2.7.14 (pip 9.0.1 from /usr/lib/python3/dist-packages (python 3.6) )
> gcc 7.2.0
> Kernel 4.13.0-16-generic

I also tried run as sudo user. Outputs the same result.

mcudev:
Same happens for me. If you comment lines 112, 121,110, and 119 in `trezor-core/src/main.py`, you'll at least get something to show up. It'll fail for another reason after that. This isn't a fix, but if you want to get rolling to start working on a fix, this'll help you to at least get something to show in the emulator.

prusnak:
Try again after pulling the master, Dominik. It should work.

dominik-selmeci:
It works! That little window is so satisfying :D.

Thanks for the help.

PR #79, opened 2017-11-06
emulator: set default HEAPSIZE to roughly equal value from firmware.elf

mcudev:
fixes https://github.com/trezor/trezor-core/issues/77

not sure why HEAPSIZE would be different based on operating system architecture, so i collapsed it. works for me.

### DIFF
--- a/emu.sh
+++ b/emu.sh
@@ -6,13 +6,8 @@ EXE=build/unix/micropython
 OPTLEVEL="${OPTLEVEL:-0}"
 MAIN="${MAIN:-main.py}"
 BROWSER="${BROWSER:-chromium}"
-if file $EXE | grep -q 80386 ; then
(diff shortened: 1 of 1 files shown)


prusnak:
Because on 64-bit build all micropython variables are twice as big and they will not fit the heap that is enough on 32-bit (x86 and ARM).

prusnak:
Fixed differently in e40c436b827080de61c0c3c353d769aa13d790e0. Thx!

mcudev:
hmm, ok, cool. i wonder if the generated binary is ever 64-bit. i'm on x86_64 and the micropython binary generated is ELF 32-bit. eh, it works.

prusnak:
It's always 32-bit, because we build it that way. Upstream build defaults to 64-bit.

PR #80, opened 2017-11-06
firmware: get_entropy text wrapping

mcudev:
testing with `trezorctl get_entropy 4` the text was going off screen.

### DIFF
--- a/src/apps/wallet/get_entropy.py
+++ b/src/apps/wallet/get_entropy.py
@@ -20,7 +20,7 @@ async def _show_entropy(ctx):
     from trezor.ui.container import Container
     from ..common.confirm import require_confirm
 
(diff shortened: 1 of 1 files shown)


mcudev:
wow, the emulator is slick!
`trezorctl -t udp get_entropy 4`

jpochyla:
Thanks! Btw, as you've started digging trough the Python code, what do think about it? It is primary meant to be a platform, so any early feedback is welcome!

mcudev:
I only know a little python right now, so I was like, "what is all this async and await stuff...I'll have to read about that sometime". :)
I ended up just trying a couple of things like load_device and wipe_device until I found something close enough that I could copy and change just a little bit for what I needed.

PR #84, opened 2017-11-12
Recovery

m1ck2:
trying to recover old wallet


prusnak:
We will pull this branch to master when it is ready.

PR #88, opened 2017-12-12
Remove PIN failure area

jhoenicke:
Instead put a small area into the norcow sector.


jhoenicke:
Added callback back to python for wait countdown

prusnak:
Thanks! :+1:

PR #90, opened 2018-01-06
Add Lisk support

alepop:
This is a work in progress to add support for `Lisk` to the `trezor-core`.
Messages: https://github.com/trezor/trezor-common/pull/68
Trezor-pyton: https://github.com/trezor/python-trezor/pull/197

Progress: 
- [x] Lisk get address
- [x] Lisk get public key
- [x] Lisk sign message
- [x] Lisk verify message
- [x] Lisk sign tx
  - [x] send transaction
  - [x] signature ransaction
  - [x] delegate transaction
  - [x] vote transaction
  - [x] multisignature transaction
_`dApp` tx types wil not allowed in the Lisk 1.0 release_
~~dapp tansaction~~
~~dapp In transfer transaction~~
~~dapp out transfer transaction~~


slush0:
Hello @alepop,any progress on this?

alepop:
@slush0 hi! Yes, I work on it, and a lot of code is not pushed to the repo. I've seen that project code is changed, and try to adapt my one to follow best practices. Why are you asking? 
I try to finish all code for the end of this week.

matejcik:
let me just point out here that `amount` and `fee` and other fields are declared as strings??
that doesn't seem right

alepop:
@matejcik Lisk is written on javascript, and as you maybe know, there is a problem with big numbers in this language. And because of this Lisk transactions use `string` for each field except `date` and `type`. Previous version use number for `amount` and `fee` but in `1.0` version will use string.

prusnak:
Well FML this is really fugly. Let me think about this. Please review
the rest and we'll discuss this later. At the moment I am inclining to
using ints in our API.

prusnak:
@alepop Can you point me to documentation where there is defined what is the valid range of values for amount/fee? I feel like we should have ints in our protocol and the conversion burden should be in the library interfacing with TREZOR, not in the TREZOR itself, so I would love to use ints in protobuf, but I need to know the range first.

alepop:
@prusnak I understand your point of view. You can't find any `valid range` for an amount in the documentation, you only can find this [64-bit integer representing the amount of Lisk to be transferred](https://lisk.io/documentation/the-lisk-protocol/transactions) in the documentation. But based on code the valid range for the amount is from `0` to [`18446744073709551615`](https://github.com/LiskHQ/lisk-js/blob/19bd6c637268c968ab0d1275bb6bc505f111a9d3/src/lisk-constants/index.js#L22).

prusnak:
@alepop last question, can you confirm that amounts and fees are always non-negative numbers? so we can use uint64?

alepop:
@prusnak yes, amounts and fees are non-negative numbers. yes, we can use`uint64`.

alepop:
@prusnak so, now I need to update `trezo-common` messages to use `uint64` and make something like converter for this fields (str -> int ) in `python-trezor`?

prusnak:
Just merged in https://github.com/trezor/trezor-common/pull/95

Please rework this PR if needed.

alepop:
@prusnak thanks, in progress

alepop:
@prusnak done.

prusnak:
@alepop thanks!

@matejcik can you please have a look at this and help with the review? @jpochyla will do the final review.

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+    # Required transaction parameters
+    t_type = pack('<b', msg.type)
+    t_timestamp = pack('<i', msg.timestamp)
+    t_amount = pack('<Q', msg.amount)
+    t_pubkey = msg.sender_public_key
+
+    if msg.requester_public_key is None:
+        t_requester_public_key = b''
+    else:
+        t_requester_public_key = msg.requester_public_key
```
matejcik: maybe here's the place for inline expression?
```python
t_requester_public_key = msg.requester_public_key or b''
```

matejcik: same below for `t_signature`, possibly similar for `t_recipient_id` ?

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if msg.recipient_id is None:
+        t_recipient_id = pack('>Q', 0)
+    else:
+        t_recipient_id = pack('>Q', int(msg.recipient_id[:-1]))
```
matejcik: I can't see the Lisk docs right now, but this is at least suspicious: all fields are packed little-endian, except for the recipient id that is big-endian?

is this the same trick as above, with the address being interpreted weirdly?
if so, there should at least be a comment explaining that this is doing address-string-to-bytes conversion

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if msg.signature is None:
+        t_signature = b''
+    else:
+        t_signature = msg.signature
+
+    t_asset = _get_asset_data_byttes(msg)
+
+    return [t_type, t_timestamp, t_pubkey, t_requester_public_key, t_recipient_id, t_amount, t_asset, t_signature]
```
matejcik: a tuple would be a better choice than a list here -- simply remove the `[]`
(at least in regular Python, unsure about uPy implementation. @jpochyla ?)

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+
+def _get_asset_data_byttes(msg):
```
matejcik: `byttes` is a typo

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+    from ustruct import pack
+    data = b''
+
+    if msg.type is Transfer and getattr(msg.asset, "data"):
```
matejcik: the `getattr` here is pointless; it will raise the same `AttributeError` that the read of `msg.asset.data` would on the following line.
if you want to check for presence of `data`, you should make that clear by saying `msg.asset.data is not None`

btw, this is not covered by the `require_confirm_by_type` call where you're catching `AttributeError`s like this.
maybe the try/catch above should extend to covert the `_get_transaction_bytes` call too?

src/apps/lisk/sign_tx.py
```diff
@@ -0,0 +1,126 @@
...
+
+async def lisk_sign_tx(ctx, msg):
+    from trezor.crypto.hashlib import sha256
+
+    public_key, seckey = await _get_keys(ctx, msg)
+    transaction = update_raw_tx(msg.transaction, public_key)
+
+    # throw ValueError if transaction has not valid structure
+    try:
+        await require_confirm_by_type(ctx, transaction)
```
matejcik: also this try/catch should maybe extend over `_get_transaction_bytes` too, because `require_confirm_by_type` does not use all relevant fields. If that's the intention. See below.

alepop:
@matejcik I appreciate your review and suggestions. Please take a look now. Also, I found a bug in logic and provide a pr with fixed tests https://github.com/trezor/python-trezor/pull/252

src/apps/lisk/helpers.py
```diff
@@ -0,0 +1,33 @@
+
+LISK_CURVE = 'ed25519'
+
+def get_address_from_public_key(public_key):
+    from trezor.crypto.hashlib import sha256
+
+    # logic from lisk-js library
+    # https://github.com/LiskHQ/lisk-js/blob/115e0e771e8456dc6c1d4acaabba93d975655cfe/lib/transactions/crypto/convert.js#L30
+    publicKeyHash = list(sha256(public_key).digest())
+    addressData = []
```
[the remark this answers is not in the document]
alepop: Thx, you are right

src/apps/lisk/verify_message.py
```diff
@@ -0,0 +1,21 @@
+
+async def lisk_verify_message(ctx, msg):
+    from trezor.crypto.curve import ed25519
+    from .helpers import get_address_from_public_key
+    from trezor import wire
+    from trezor.messages.Success import Success
+    from trezor.messages.FailureType import ProcessError
+    from apps.wallet.verify_message import require_confirm_verify_message
+
+    # Lisk signature can be more than 64 bytes
```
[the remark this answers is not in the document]
alepop: Lisk use `js-nacl` library for calculating signature, and for message signing use this method https://github.com/tonyg/js-nacl/blob/334f0587103bd3b7e721b92fc1d2a38a8f23708d/README.md#naclcrypto_signmsgbin-signersecretkey--uint8array and final signature more than 64 bytes, because contains a message body.  trezor-core `d25519.verify` raise an error if a signature is not equal 64 bytes.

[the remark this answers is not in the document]
matejcik: okay, but the protobuf message has a separate field for message and signature.
ISTM we should either take "signed mesage" straight from NaCl and extract the message from that, instead of using a separate field. (and also generate the same thing)
Or the caller (python-trezor / wallet) should do this job and pass the signature separately.

Definitely not "sig + msg" and "msg" as two separate fields. I mean, I don't see any actual security problem with this, but  it's a weird looking cracks for things to fall through; you're throwing away part of a signature without even looking at what's there.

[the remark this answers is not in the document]
alepop: @matejcik I've looked at the upcoming `1.0` version of Lisk source code now and found that they will use `crypto_sign_detached` instead of `crypto_sign` like in current version for message signing, and signature will not have a msg body. My bad, I missed this part. So, I will update this PR and make new for `python-trezor`

jpochyla:
Thank you for promptly responding @alepop, but after further discussions, I'm a bit hesitant to merge the sign_message/verify_message code. Signing arbitrary digests with arbitrary key vector is dangerous. Let's escalate this up stream. In case more restricted signing scheme gets adopted, we can revisit it later.

So, basically -- ACK with a bit of style changes TBD on my part, just without the sign/verify code.

alepop:
@jpochyla sorry I don't understand what part of sign/verify message bothered you? I fixed the part of code when I take only `64` bytes of a long signature because now Lisk uses `crypto_sign_detached` and computed signature are exactly `64` bytes long because don't include message body anymore.  Thanks.

jpochyla:
@alepop In Bitcoin and clones, before signing a message, we prefix it with a header:
https://github.com/trezor/trezor-core/blob/master/src/apps/common/signverify.py#L12

Reasons are:

1. separation of message signing and transaction signing. When I'm signing arbitrary Bitcoin message, I can be sure that it won't lead to coinloss. `apps.lisk` signs whatever you give to it, even a transaction digest.
1. separations of coins on the same curve. For example Litecoin has a different message prefix. `apps.lisk` doesn't do that, it allows signing any message with any key derived from seed on ed25519 curve.

alepop:
@jpochyla got it. thank you for the explanation

prusnak:
JFYI: I removed `Lisk{SignMessage,MessageSignature,VerifyMessage}` in https://github.com/trezor/trezor-common/commit/fb662e53b129b250149d982259696531ff4c2b95, so let's drop Sign/Verify message for Lisk for now.

alepop:
@prusnak @jpochyla I've opened an [issue](https://github.com/LiskHQ/lisk-elements/issues/673) about signed message prefix in the LiskHQ repository. I will make new pr with message sign/verify after correct implementation for this feature in Lisk.

alepop:
for the record, open pr with restored functionality for signing and verifying messages https://github.com/trezor/trezor-core/pull/217

PR #92, opened 2018-01-11
Ethereum signing

tsusanka:
Some final fixes for Ethereum. Few tiny things are missing:

- check the "ux" and wording in layout.py
- data progress is not implemented


prusnak:
Cool!

Added commit af93c7299dea6cbf64f90c7180de8ff86332738c which fixes compatibility with T1 and also passes the device test in python-trezor.

PR #94, opened 2018-01-17
embed/unix/usb.c: usb write fix

ph4r05:
Hi!

I've got some small assertion problem with `usb.c`. This PR fixes it. 

Pls take a look.

The problem:

```
FATAL ERROR:
expr: usb_hid_write(0, (const uint8_t *)ping_resp, strlen(ping_resp))
msg : usb_hid_write
file: embed/unix/usb.c:121
func: usb_hid_read
rev : 9c71bd8-dirty
SHUTDOWN
```


prusnak:
Thanks for the fix. I fixed the issue a little bit differently in 3098dc8447850ef1e08ca61aecb6e37ae2bea1b0 (didn't like we are misusing usb_hid_write from usb_hid_read).

ph4r05:
ok cool! thanks for the fix.

PR #100, opened 2018-01-23
Terracoin: Add network and test

TheSin-:
Adding Terracoin (TRC) to Trezor, this adds the coin and adds it to the tests.

This compliments the PR on trezor-common https://github.com/trezor/trezor-common/pull/75


prusnak:
No need to send a separate PR here. The coin definition will get propagated from trezor-common repository.

PR #106, opened 2018-02-08
webusb: avoid chrome http bug and nice side-effect of avoiding annoying notifications

mcudev:
I was testing with the latest version of Chrome on a Ubuntu machine with Wireshark running.

When I plugged the device in, Chrome showed a notification. When I clicked on that notification, Chrome erroneously generated a plain HTTP request to http://trezor.io/start. This is a Chrome bug as it ignored the descriptor value being set to the HTTPS scheme (I verified the descriptor value with Wireshark listening on my USB bus also) (descriptor reference https://wicg.github.io/webusb/#webusb-descriptors). This bug is tied to Chrome's current cache status as well. So, clear your cache to easily reproduce.

This PR sets the scheme value to 255 which is supposed to encode the entire URL. This causes Chrome to neither show the notification, nor send a request. I'm fine with this because I thought that the notification was annoying anyways and at least it doesn't cause insecure requests to be processed by the browser.

### DIFF
--- a/embed/bootloader/bootui.c
+++ b/embed/bootloader/bootui.c
@@ -111,7 +111,8 @@ void ui_screen_third(void)
 {
     display_bar(0, 0, DISPLAY_RESX, DISPLAY_RESY, COLOR_WHITE);
     display_icon((DISPLAY_RESX - 180) / 2, (DISPLAY_RESY - 30) / 2, 180, 30, toi_icon_welcome + 12, sizeof(toi_icon_welcome) - 12, COLOR_BLACK, COLOR_WHITE);
-    display_text_center(120, 220, "Go to trezor.io/start", -1, FONT_NORMAL, COLOR_BLACK, COLOR_WHITE);
+    display_text_center(120, 200, "Go to", -1, FONT_NORMAL, COLOR_BLACK, COLOR_WHITE);
+    display_text_center(120, 220, "https://trezor.io/start", -1, FONT_NORMAL, COLOR_BLACK, COLOR_WHITE);
 }
 
 // info UI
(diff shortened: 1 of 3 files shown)


mcudev:
Also, if this PR is accepted, then probably something similar for v1 devices should be done.

prusnak:
Nice find, but I would rather not workaround this in our stack. The bug will be fixed in Chrome eventually and we have a redirect from HTTP to HTTPS.

mcudev:
Yes, I saw the 301 redirect. It leaves some to be desired from a SSL stripping and response manipulation perspective. This is kind of a pet peeve of mine. I was also going to work on a commit to change all the trezor.io/start URLs to https. When I help people setup devices, I always have to tell them to put "https://" in front of the URL on the device screen. I get that the overall security model can withstand a malicious website (due to on device verification), but by using http sometimes, it reduces the barrier to user error causing loss.

mcudev:
The nice side effect of this PR stopping the trezor.io/start notificationon every plug when Chrome is active is because that notification only helps new users. After initialization, it would be more helpful to notify a link to https://wallet.trezor.io/. I think the notifications are annnoying all the time. But that's just me :)

mcudev:
Related; the redirects could use some updating too since the trailing slash matters:
http://trezor.io/start -> https://trezor.io/start
**https**://trezor.io/start -> **http**://trezor.io/start/
http://trezor.io/start/ -> https://trezor.io/start/

prusnak:
I see your points and kind of agree with them, but like you said, I am going to close this unmerged. 

We purposefully copied the narrative from Chrome popup and used "Go to trezor.io/start" so the user is not confused. Also I suppose that Chrome is [slowly trying to kill HTTP](https://blog.chromium.org/2018/02/a-secure-web-is-here-to-stay.html), so we should be good eventually.

Thanks for the ideas, nevertheless!

mcudev:
@prusnak hey, i'm stupid, i just looked closer at the network capture. the http request is not a chrome bug. it is a result of the weird redirects that i mentioned where `https://trezor.io/start` redirects to `http://trezor.io/start/` which then redirects back to `https://trezor.io/start/`.
easiest fix is to fix the redirect loop by making `https://trezor.io/start` redirect to `https://trezor.io/start/`.

prusnak:
Ah, I remember. This is caused by a combination of Clouflare running on
top of GitHub pages.
GitHub does the redirect from /start to /start/ but it uses HTTP,
because it doesn't now about Cloudflare running on top.
This could be fixed in CF and we are looking into it ATM.

slush0:
Fixed now, thanks for pointing it out.

mcudev:
yep, i confirm that fixed it. `https://trezor.io/start` returns a 302 instead of 301, but that is very minor.

PR #108, opened 2018-02-10
bootloader: fix process_msg_unknown remaining chunk calculation

mcudev:
I was getting a red screen in the bootloader when I sent the unknown/invalid message `reset_device`. The message size, 22 was less than 55, and thus wrapping around to a large unsigned int. Also, the calculation was not rounding up to process all bytes. Finally, I made a small change to actually use the `iface_num` argument. I have not checked to see if a similar bug is in the firmware, but I'll look and open a separate PR if I find one.

I verified that this patch works against these tests that send messages having 1, 2, and 3 chunks:

```
python2 trezorctl reset_device

python2 trezorctl load_device -m 'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about'

python2 trezorctl load_device -m 'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon agent'
```

### DIFF
--- a/embed/bootloader/messages.c
+++ b/embed/bootloader/messages.c
@@ -553,9 +553,15 @@ int process_msg_WipeDevice(uint8_t iface_num, uint32_t msg_size, uint8_t *buf)
 void process_msg_unknown(uint8_t iface_num, uint32_t msg_size, uint8_t *buf)
 {
     // consume remaining message
-    int remaining_chunks = (msg_size - (USB_PACKET_SIZE - MSG_HEADER1_LEN)) / (USB_PACKET_SIZE - MSG_HEADER2_LEN);
+    int remaining_chunks = 0;
+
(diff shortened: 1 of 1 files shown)


prusnak:
I know that `bootloader_usb_loop` eats the remaining bytes, but I like that we have a specific eater in `process_msg_unknown`. Therefore I cherry-picked just the first commit and left out the second one. Thanks!

prusnak:
The second wrong usage of  USB_IFACE_NUM was fixed in 8a409bd99a79fe8861dee50dfacae8de3e117f35

mcudev:
the only other thing that i thought of that makes me like the second commit more is that if the msg_size is specified to be like 0xffffffff, then the calc does a -55 + 62 and wraps around. we fall back to the `bootloader_usb_loop` eater anyways. i don't think we can trust the msg_size, it seems like more of a hint for what the size might be.

PR #110, opened 2018-02-10
firmware: fix bugs affecting pb message handling and wipe device flow

mcudev:
fix bugs introduced by c1e1e8bf0255abc1d4de1dd0e6541b139dc7b0c6

### DIFF
--- a/src/apps/common/cache.py
+++ b/src/apps/common/cache.py
@@ -22,4 +22,4 @@ def set_seed(seed):
 
 def clear():
     global _seed, _state
-    _seed, _state = None
+    _seed, _state = None, None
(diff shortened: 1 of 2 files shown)


prusnak:
The second fix should not be needed IMHO. If the field is not present, the parse should create a None entry.

Have you pulled in this change? https://github.com/trezor/trezor-core/commit/6fad2f4283f25fa280b0606c570a4f941cbe84a4#diff-8d77dcd85cebc83082ad5ffbcc3bc7b1 (it adds state field into the Initialize structure).

mcudev:
yep, i was testing against commit c1e1e8bf0255abc1d4de1dd0e6541b139dc7b0c6
more info here https://github.com/trezor/trezor-core/commit/c1e1e8bf0255abc1d4de1dd0e6541b139dc7b0c6#r27476831

mcudev:
yeah, you're right, just take the one change. closing.

PR #117, opened 2018-02-17
Use GCC as linker for firmware build

saleemrashid:
Using GCC passes the correct flags to the linker. Including, if we choose to use features such as LTO, linker plugins.

Needs to be tested on real hardware because it gives very minor differences in the binary.

### DIFF
--- a/SConscript.firmware
+++ b/SConscript.firmware
@@ -297,7 +297,7 @@ env.Replace(
     AS='arm-none-eabi-as',
     AR='arm-none-eabi-ar',
     CC='arm-none-eabi-gcc',
-    LINK='arm-none-eabi-ld',
+    LINK='arm-none-eabi-gcc',
     SIZE='arm-none-eabi-size',
     STRIP='arm-none-eabi-strip',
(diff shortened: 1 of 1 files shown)


SConscript.firmware
```diff
@@ -314,7 +314,7 @@ env.Replace(
     '-fstack-protector-all '
     + CCFLAGS_MOD,
     CCFLAGS_QSTR='-DNO_QSTR -DN_X64 -DN_X86 -DN_THUMB',
-    LINKFLAGS='-nostdlib -T embed/firmware/memory.ld --gc-sections -Map=build/firmware/firmware.map --warn-common',
```
prusnak: Why have you removed `-nostdlib`?

saleemrashid: It's included in `CCFLAGS`

prusnak:
Applied as bigger changeset (for all targets) here: 3f63acdd7751c91c3814dc066b9b748952549c8d

Also I tested that bootloader built using previous method works with new firmware and also that new boardloader + bootloader + firmware work together.

prusnak:
Thanks!

PR #119, opened 2018-02-19
Viacoin: Add network and test

romanornr:


prusnak:
Please submit a pull request to https://github.com/trezor/trezor-common/blob/master/coins.json where we hold coin definitions. The python file you edited is generated from JSON source.

PR #130, opened 2018-02-28
Making bigger heapsize

karel-3d:
With current heapsize, emulator constantly fails; I made it 800k and it worked.

It is not scientific, I just increased the number and webwallet (with trezord-go from master) started to work.


prusnak:
Fixed in b77646904b172f6c639650e5639a4fc051982537

PR #134, opened 2018-03-01
embed/unix: emit TOUCH_END when mouse leaves active area.

matejcik:
This fixes the "swiping bug", swiping in emulator now feels more natural.


prusnak:
:+!

PR #140, opened 2018-03-04
Wanchain support in trezor-core

tyrion70:
Added wanchain_sign_tx function and WanchainSignTx message. Tried to do it with as less duplication as possible. Because there is an extra field in the beginning of the message (txtype) I couldn't modify EthereumSignTx and had to create a new message and some functions.


jhoenicke:
Please note that the files in `trezorlib.messages` are generated from the protobuf definitions in `trezor-common/protob`.  You should change [messages.proto](https://github.com/trezor/trezor-common/blob/master/protob/messages.proto) and then regenerate the files.

You can add the new txtype to ethereum_sign_tx; just give it a new number and mark the field as optional.  The order of the fields doesn't have to match the order in which they are signed.  The function could then check if txtype is present (check for None).

Does wanchain uses a different chain_id?  We usually use the chain_id to determine the currency that is displayed on the confirmation screen.

tyrion70:
Unfortunately Wanchain does not have a different chain_id (yet). I think they didn't change it because adding an extra field ensures replay protection in itself, but I expect it to be changed pretty soon. I'll build the stuff with protobuf (had trouble getting that working earlier so I'll give it another go). That does mean I need to make a PR to trezor-common as well btw.

prusnak:
Closing this. Please reopen with new changeset once we have merged the trezor-common change (ideally minimal required change to get this working).

PR #148, opened 2018-03-06
passphrase: include all ASCII characters

jhoenicke:
I'm not sure about the optimal order.  Also note that `*` and `#` are missing since they are already on the alphabet layers.

passphrase


prusnak:
Thank you, Jochen! :+1:

PR #149, opened 2018-03-06
Wanchain

tyrion70:
Added support for Wanchain. In order to display the name Wan in the display I made a change to layout.py and networks.py since the chain_id of wanchain is currently set to 1 as well. If and when the chain_id is set to another id it can be reverted. Hope this is an acceptable solution for now.

### DIFF
--- a/src/apps/ethereum/layout.py
+++ b/src/apps/ethereum/layout.py
@@ -8,26 +8,26 @@ from . import networks
 from .get_address import _ethereum_address_hex
 
 
-async def require_confirm_tx(ctx, to, value, chain_id, token=None):
+async def require_confirm_tx(ctx, to, value, chain_id, token=None, tx_type=None):
     if to:
         str_to = _ethereum_address_hex(to)
     else:
         str_to = 'new contract?'
     content = Text('Confirm sending', ui.ICON_SEND,
-                   ui.BOLD, format_ethereum_amount(value, token, chain_id),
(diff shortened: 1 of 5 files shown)


jhoenicke:
I think we should check that tx_type is None or 1.  Just as a safety precaution that we don't sign a transaction we don't understand. 

Is there a reason to use "bytes" for tx_types instead of, e.g., uint32?  The latter would be easier, but may be an issue if they use tx_types with more than 32 bits some day.

tyrion70:
I checked about the tx_type, its currently either 0x01 or 0x06 so I can change it to an uint32 safely. I agree that adding a precaution to sign only those is prudent, I'll add that to the check function in sign_tx.py. 

Regarding the tx_type change to uint, I can do that, but I'd need to change trezor-common then as well again. @prusnak you got any preferences here? I would personally prefer keeping it bytes now since it would have been easier only if I had done it from the start ;) But I'm fine with changing it in all repos.

prusnak:
Please change to uint32 if possible.

tyrion70:
Tx_type was changed to uint in all three repo's. latest trezor-common was used to generate again. Check was also added to only allow tx_type 1, 6 and None.
@jhoenicke @prusnak please let me know if i need to add anything more. Thanks!

prusnak:
Can you please rebase on top of master? There were some fixes and your changes are in conflict with them.

tyrion70:
@prusnak I just rebased with master as per your request.

prusnak:
Thanks!

PR #150, opened 2018-03-08
Ethereum: Fix fee computation.

jhoenicke:
- Gas is always in ether, even when sending tokens.
- Fee is computed by multiplying gas limit with gas price.
- Parse numbers already in sign_tx.
- Made rlp.encode non-recursive (also fixes not passing include_length).


prusnak:
Cc @tsusanka

tsusanka:
Ah, I expected the gas limit already includes the gas price. Thanks!

prusnak:
:+1

PR #156, opened 2018-03-09
Added Bitcoin Private to coins

csulmone:


jpochyla:
> `# the following list is generated using tools/codegen/gen_coins.py`
> `# do not edit manually!`

Coins are defined in https://github.com/trezor/trezor-common repository.

PR #158, opened 2018-03-09
apps/wallet/ecdh: support decryption (for GnuPG)

romanz:
Tested both with NIST256 and Curve25519 GnuPG keys:

screenshot at 2018-03-09 17-53-36

(similar to https://github.com/trezor/trezor-mcu/commit/c86086e2b90e0a2e547d661766e05585d2cf54f5)

### DIFF
--- a/src/apps/wallet/__init__.py
+++ b/src/apps/wallet/__init__.py
@@ -6,6 +6,7 @@ from trezor.messages.wire_types import \
     SignTx, \
     SignMessage, VerifyMessage, \
     SignIdentity, \
+    GetECDHSessionKey, \
     CipherKeyValue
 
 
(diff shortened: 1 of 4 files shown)


prusnak:
Can you please add GetECDHSessionKey unit test to https://github.com/trezor/python-trezor/tree/master/trezorlib/tests/device_tests similar to https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_getpublickey_curve.py?

So we can test behaviour on both T1 and T2 and whether they are the same?

romanz:
Sounds good, will add it today :)

prusnak:
Thank you!

PR #178, opened 2018-03-26
Allow registering custom protobuf messages

slush0:
New applications should self-contain protobuf messages, instead of storing them in global src/trezor/messages namespace.


jpochyla:
Reimplemented with little changes in 60bec0b4d1887579659bf1a31f24cb3cbf89cb34 and 2731742d340429fff7aa1ef401357d1b98d3dff5, thank you!

PR #182, opened 2018-04-04
vendor: trezor-crypto updated to b0af159

tsusanka:
@jpochyla may I? I could use the nem related commits


prusnak:
vendor/trezor-crypto updated in master. Closing

PR #185, opened 2018-04-05
src/trezor/wire: add exceptions for all defined FailureTypes

jpochyla:
Makes the error API much more ergonomic.


prusnak:
utACK

PR #191, opened 2018-04-16
replace mp_obj_get_int_truncated with mp_obj_int_get_checked

peter-conalgo:
I strongly recommend you change `mp_obj_get_int_truncated` into `mp_obj_int_get_checked` in
`modtrezorcrypto-random.h` [on line 29.](https://github.com/trezor/trezor-core/blob/14b0d4439a2f348d3e25ab2b5a4e01babe6363dd/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h#L29)

Because otherwise, `modtrezorcrypto.random.uniform((2*32)+1)` will always return zero.

Similarly, any call with an argument larger than 2**32 will not do what you want.

I couldn't find any codepaths where that could happen under control of an attacker, but it's quite possible there is one, or could be someday.

### DIFF
--- a/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h
+++ b/embed/extmod/modtrezorcrypto/modtrezorcrypto-random.h
@@ -26,7 +26,7 @@
 ///     Compute uniform random number from interval 0 ... n - 1.
 ///     '''
 STATIC mp_obj_t mod_trezorcrypto_random_uniform(mp_obj_t n) {
-    uint32_t nn = mp_obj_get_int_truncated(n);
+    uint32_t nn = mp_obj_int_get_checked(n);
     if (nn == 0) {
         mp_raise_ValueError("Maximum can't be zero");
     }


prusnak:
We also use `mp_obj_get_int_truncated` in two more locations:

1) BIP32: https://github.com/trezor/trezor-core/blob/3b67cb6bea3f4a52b4b531c5fe3d54c510831b55/embed/extmod/modtrezorcrypto/modtrezorcrypto-bip32.h#L69

2) CRC: https://github.com/trezor/trezor-core/blob/3b67cb6bea3f4a52b4b531c5fe3d54c510831b55/embed/extmod/modtrezorcrypto/modtrezorcrypto-crc.h#L26

I would rather change all 3 usages or neither.

prusnak:
@jpochyla thoughts?

jpochyla:
> I would rather change all 3 usages or neither (while changing the documentation to mention the limitation) because they are all semantically the same (user wants to pass uint32 value from python to C code).

I agree, let's change it to `mp_obj_get_int` everywhere.  That should:

1. return good error message for values of incorrect type
2. return good error message for overflowing values

One possible down-side is that boolean values would get automatically converted to int, but I think we can live with that.

> We also use mp_obj_get_int_truncated at two more locations:

Also `mp_obj_int_get_truncated` should get replaced:

https://github.com/trezor/trezor-core/blob/3b67cb6bea3f4a52b4b531c5fe3d54c510831b55/embed/extmod/modtrezorio/modtrezorio-poll.h#L60

jpochyla:
Superseded by #201.

PR #193, opened 2018-04-20
Dash Testnet support

Bertrand256:
Please, enable Dash Testnet support.


prusnak:
Coins should be added to https://github.com/trezor/trezor-common/blob/master/coins.json not to this generated file.

But we won't be adding testnet coins to the production firmware anyway, so don't send a PR there.

Bertrand256:
Oops, I somehow missed that it was a generated file.
I see, however, that several testnets are supported in the production firmware. 

Well, we will have to live with it using the not-so-comfortable solution based on custom firmware. Anyway, thanks for the quick review.

Bertrand256:
Oops, I somehow missed that it was a generated file.
I see, however, that several testnets are supported in the production firmware. 

Well, we will have to live with it using the not-so-comfortable solution based on custom firmware. Anyway, thanks for the quick review.

PR #195, opened 2018-05-04
use pipenv to run python-trezor testsuite reproducibly

matejcik:
this freezes dependency versions (most importantly right now, `rlp` is frozen at 0.6.0 which is the last that supports Travis python3.4) so that third-party changes shouldn't break the builds anymore.

Along with some nice new docs about testing.


prusnak:
LGTM! :+1:

PR #196, opened 2018-05-09
protobuf: support signed ints properly

### LINKED ISSUE #189: NEM: removing a cosignatory fails
When modifying the number of cosignatories, the number is not interpreted properly. Might be fixed by https://github.com/trezor/python-trezor/issues/249 but needs to be verified when merged

matejcik:
this is an equivalent of https://github.com/trezor/python-trezor/pull/250 for core
it should fix #189

### DIFF
--- a/src/protobuf.py
+++ b/src/protobuf.py
@@ -40,6 +40,8 @@ async def load_uvarint(reader):
 
 
 async def dump_uvarint(writer, n):
+    if n < 0:
+        raise ValueError("Cannot dump signed value, convert it to unsigned first.")
     buffer = _UVARINT_BUFFER
     shifted = True
     while shifted:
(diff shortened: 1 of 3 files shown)


src/protobuf.py
```diff
@@ -49,15 +51,45 @@
...
+# Then you can take "number XOR all-sign-bits", which is XOR 0 (identity) for positive
+# and XOR 1 (bitwise negation) for negative. Cute and efficient.
+#
+# But this is harder in Python because we don't natively know the bit size of the number.
+# So we have to branch on whether the number is negative.
+
+def sint_to_uint(sint):
+    res = sint << 1
+    if sint < 0:
+        res = ~res
```
prusnak: Will this work correctly for both sizes (32 and 64 bit?).

I would love to see some unit tests for these.

src/protobuf.py
```diff
@@ -49,15 +51,45 @@
...
+    return res
+
+
+def uint_to_sint(uint):
+    sign = uint & 1
+    res = uint >> 1
+    if sign:
+        res = ~res
```
prusnak: Same as above

src/protobuf.py
prusnak: If there are tests in python-trezor, then that's enough. Thanks!

tools/pb2py
```diff
@@ -38,14 +38,10 @@ def process_type(t, cls, msg_id, indexfile, is_upy):
             type = 'p.UVarintType'
             pytype = 'int'
 
-        elif v.type in (17,):
+        elif v.type in (17,18,):
```
prusnak: Please add a space after the first comma. Maybe also add tools/pb2py and other python tools that do not end with .py to `make style` target so flake8 will test them.

prusnak:
utACK

src/protobuf.py
matejcik: here are the tests, for the record
https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/unit_tests/test_protobuf.py

PR #197, opened 2018-05-09
pb2py refactor

matejcik:
basically the goal here is to have an identical `build_protobuf` and `pb2py` between trezor-core and python-trezor

i took the opportunity to make better type hints, fix the kwargs syntax error that happens to work in uPy but not CPy, and some other minor things.


tools/pb2py
```diff
@@ -2,239 +2,213 @@
...
+    fields = list(parse_field(number, field)
+                  for number, field
+                  in descriptor.fields_by_number.items())
+
+    if any(field.repeated for field in fields):
+        yield "if __debug__:"
+        yield "    try:"
+        yield "        from typing import List"
+        yield "    except ImportError:"
+        yield "        List = None"
```
matejcik: an interesting side-effect of this particular choice (`List = None` as opposed to e.g. `List = {}`) is that it caught CPython redhanded at processing type hints but not having a `typing` module.

that might turn out to be a problem in some circumstances, but maybe it's better to have it?

prusnak:
Merged via 5885cb867e0de745ae3fc7e701248ecaa4fc64a3

matejcik:
note that this will break without #196

PR #200, opened 2018-05-10
eth: unknown token transfer is recognized and displayed as unknown

### LINKED ISSUE #198: T2: Invalid amount and token name while sending Erc20 tokens
signing erc20 transaction will display amount of tokens in Wei as "eth" instead of "token_name"

tsusanka:
When unknown token was transferred (meaning: the token address was not listed in the tokens list), we didn't treat it as a token transfer but as some arbitrary data. Since we know it is a token transfer it is beneficial to display the user such information and notify them that the token is unknown. In other words that the value is unknown, because we do not know the divisibility, nor the token suffix.

closes #198

### DIFF
--- a/src/apps/ethereum/tokens.py
+++ b/src/apps/ethereum/tokens.py
@@ -1,11 +1,11 @@
 def token_by_chain_address(chain_id, address):
-    if not address:
-        return None
     for token in tokens:
         if chain_id == token[0] and address == token[1]:
             return token
-    return None
+    return UNKNOWN_TOKEN
 
+
+UNKNOWN_TOKEN = True
 
 # rest of the file is generated using trezor-common/ethereum_tokens-gen.py
 # DO NOT EDIT MANUALLY!
--- a/src/apps/ethereum/tokens.py
+++ b/src/apps/ethereum/tokens.py
@@ -5,7 +5,7 @@ def token_by_chain_address(chain_id, address):
     return UNKNOWN_TOKEN
 
 
-UNKNOWN_TOKEN = True
+UNKNOWN_TOKEN = (None, None, None, None)
 
 # rest of the file is generated using trezor-common/ethereum_tokens-gen.py
 # DO NOT EDIT MANUALLY!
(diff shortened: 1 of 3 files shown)


src/apps/ethereum/tokens.py
```diff
@@ -1,12 +1,12 @@
 def token_by_chain_address(chain_id, address):
-    if not address:
-        return None
     for token in tokens:
         if chain_id == token[0] and address == token[1]:
             return token
     return None
 
 
+UNKNOWN_TOKEN = True
```
tsusanka: @jpochyla I wans't sure to what value I should declare this :-/ . Also I've forgotten `const()` haven't I?

prusnak:
I would change `token_by_chain_address` to return UNKNOWN_TOKEN and drop
conversion from None to UNKNOWN_TOKEN in format function.
Also not sure about using True as UNKNOWN_TOKEN, seems a little bit
hackish. @matejcik what do you think?
Thirdly, const(foo) makes sense only for int values, not bool.

prusnak:
I meant conversion in sign, not format.

tsusanka:
> Also not sure about using True as UNKNOWN_TOKEN, seems a little bit
hackish. @matejcik what do you think?

Yeah, agree.

> Thirdly, const(foo) makes sense only for int values, not bool.
I was thinking about `const(1)` but it probably isn't better

src/apps/ethereum/tokens.py
matejcik: I suggest `UNKNOWN_TOKEN = object()`

tests/test_apps.ethereum.tokens.py
```diff
@@ -22,7 +22,7 @@ def test_token_by_chain_address(self):
 
         # invalid adress, invalid chain
         token = tokens.token_by_chain_address(999, b'\x00\xFF')
-        self.assertEqual(token, None)
+        self.assertEqual(token, tokens.UNKNOWN_TOKEN)
```
matejcik: this should probably be `assertIs`

tsusanka:
I think we can merge this, agree?

prusnak:
utACK

tsusanka:
I'm daring to merge this, it's pretty small. cc @jpochyla have a look when you come back, just so you know

PR #206, opened 2018-05-24
wallet: isinstance in __init__

tsusanka:
This is just a tiny refactoring of the init function in the wallet, `isinstance` is now used. Tests are passing, so good to merge?


tsusanka:
I've found just one more instance of this. Should be all

PR #207, opened 2018-05-25
crypto: base32

tsusanka:
I need base32 for Stellar, but I'm filling this as a separate PR to make it a bit clearer.

I've copied the base32 implementation from the [micropython lib](https://github.com/micropython/micropython-lib/blob/master/base64/base64.py) with one minor modification: the `encode` function returns _str_ instead of _bytes_. See [line 200](https://github.com/micropython/micropython-lib/blob/master/base64/base64.py#L200) vs [66](https://github.com/trezor/trezor-core/compare/tsusanka/base32?expand=1#diff-f30d492a9c1f9b0ba18a206143d41af2R66). I'm now looking at the [official Python base32](https://github.com/python/cpython/blob/c643a967dd7d33ccefa5b61b38caf40b448057ce/Lib/base64.py#L142) implementation though and it returns bytes as well, so maybe this is a bad idea?

### DIFF
--- a/src/trezor/crypto/base32.py
+++ b/src/trezor/crypto/base32.py
@@ -1,0 +1,150 @@
+# Base32 implementation taken from the micropython-lib's base64 module
+# https://github.com/micropython/micropython-lib/blob/master/base64/base64.py
+#
+
+import ubinascii
+import ustruct
+
+bytes_types = (bytes, bytearray)  # Types acceptable as binary data
+
+_b32alphabet = {
+    0: b'A',  9: b'J', 18: b'S', 27: b'3',
+    1: b'B', 10: b'K', 19: b'T', 28: b'4',
+    2: b'C', 11: b'L', 20: b'U', 29: b'5',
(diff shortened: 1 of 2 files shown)


matejcik:
other `encode` functions (base64, binascii.hexlify) also return bytes. I wonder what's the rationale for that.

matejcik:
ah, here it is: https://stackoverflow.com/a/42776711/222189
Python's stdlib considers "bytes -> str" a _decode_ operation, but in this case it would be an _encode_. So it's an API regularity thing.
That reasoning doesn't apply to us. But it might be a matter of conforming to Pythonist's expectations?

prusnak:
utACK

tsusanka:
> That reasoning doesn't apply to us. But it might be a matter of conforming to Pythonist's expectations?

Yeah, I think I agree, I guess it's better to remove my modification. Let's see what @jpochyla thinks though

prusnak:
Merged via bb57000449bae63f12e5c3dbc720802df4850fb6 with small changes

PR #213, opened 2018-06-04
Efficient text layout

jpochyla:
First shot at the algorithm.

### DIFF
--- a/src/trezor/ui/text.py
+++ b/src/trezor/ui/text.py
@@ -4,36 +4,110 @@ from trezor import ui
 TEXT_HEADER_HEIGHT = const(48)
 TEXT_LINE_HEIGHT = const(26)
 TEXT_MARGIN_LEFT = const(14)
+TEXT_MAX_LINES = const(4)
+
+BR = const(-1)
+
+
+def render_words(words: list, new_lines: bool, max_lines: int) -> None:
+    # initial rendering state
+    font = ui.NORMAL
+    fg = ui.FG
+    bg = ui.BG
+    offset_x = TEXT_MARGIN_LEFT
+    offset_y = TEXT_HEADER_HEIGHT + TEXT_LINE_HEIGHT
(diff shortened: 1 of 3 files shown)


src/trezor/ui/text.py
```diff
@@ -4,36 +4,109 @@
...
+    for word in words:
+        if isinstance(word, int):
+            if word == BR:
+                # line break
+                if not offset_y < OFFSET_Y_MAX:
+                    ui.display.text(offset_x, offset_y, '...', ui.BOLD, ui.GREY, bg)
+                    return
+                offset_x = TEXT_MARGIN_LEFT
+                offset_y += TEXT_LINE_HEIGHT
+            elif word == ui.NORMAL or word == ui.BOLD or word == ui.MONO:
```
prusnak: This would not work as expected, because `ui.NORMAL == ui.BLACK` see https://github.com/trezor/trezor-core/blob/master/src/trezor/ui/style.py#L36 and https://github.com/trezor/trezor-core/blob/f72667c761e426bba229441aa65d934be53a8af3/embed/extmod/modtrezorui/display.h#L35-L43

Should we use different constants (bigger than 16 bits, e.g. 0x10000, 0x20000, 0x30000, or negative, e.g. -1, -2, -3) for ui.NORMAL, ui.BOLD and ui.MONO?

If yes, we'd need to perform a conversion here: https://github.com/trezor/trezor-core/blob/f72667c761e426bba229441aa65d934be53a8af3/embed/extmod/modtrezorui/modtrezorui-display.h#L207 from uPy constant to display.h constant.

jpochyla:
Good catch! BR is defined as -1 now.

________________________________

src/trezor/ui/text.py
jpochyla: Fixed in 2b770093c79d33a505d341d80040c1dd45a96932

PR #222, opened 2018-06-08
Crown Support

Extrememist:
Addition of Crown


prusnak:
We'll process the coin addition manually.

prusnak:
I just noticed that `address_type=0` - we won't be adding this coin unless you change address_type of the currency so it does not collide with Bitcoin.

Extrememist:
Heya, Thanks for getting back about that :) will try have something done

PR #230, opened 2018-06-11
bootloader: ask again for block if invalid

### LINKED ISSUE #135: Bootloader: ask again for block if invalid
If invalid block is encountered during firmware update, the whole operation fails. Bootloader should try to redownload the block again (3-times maybe?).

malinah:
fixes #135


prusnak:
LGTM Thanks!

PR #231, opened 2018-06-11
Update trezor-common

ilanolkies:
### General
Rootstock is no more the network name, it's **RSK**.

### Related
- [trezor-common issue](https://github.com/trezor/trezor-common/issues/130).

### Solution
Update trezor-common submodule.


prusnak:
Will be fixed as a part of bigger fix

ilanolkies:
Thank you! This change also applies for trezor-mcu.

PR #232, opened 2018-06-11
RSK uses a different checksum encoding.

ilanolkies:
### General
RSK uses a different address checksum encoding, not the same as Etherum. It's described in RSKIP-60, and it's compatible with EIP-55 ecnoding.

### Solution
- Implement RSK checksum in Ethereum `require_confirm_tx`
- Implement RSK checksum in `EthereumGetAddress`
     1. Derive node id from Derivation Path
     2. Apply checksum

### DIFF
--- a/src/apps/ethereum/get_address.py
+++ b/src/apps/ethereum/get_address.py
@@ -26,12 +26,15 @@ async def ethereum_get_address(ctx, msg):
     return EthereumAddress(address=address)
 
 
-def _ethereum_address_hex(address):
+def _ethereum_address_hex(address, chain_id=None):
     from ubinascii import hexlify
     from trezor.crypto.hashlib import sha3_256
 
+    applying_chain_ids = [30, 31]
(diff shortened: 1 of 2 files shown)


prusnak:
Please squash your commits into one. Also please rebase on top of the master and use the new `networks` api: https://github.com/trezor/trezor-core/commit/ad1e0e81b227188e660c6567fd6e009d14007bba (use `networks.by_chain_id` or `networks.by_slip44` methods to retrieve the `NetworkInfo` class and check whether its field rskip60 is set to `True` or `False).

ilanolkies:
Hi! @prusnak . EthereumGetAddress receives an "unsigned int 32" representing slip-44 coin id. Do you have any function to "parse it to int"?

prusnak:
You can use `val & 0x7fffffff`

ilanolkies:
True! Thanks!

PR #235, opened 2018-06-12
RSKIP-60 based checksum address encoding

ilanolkies:
### General
RSK uses a different address checksum encoding, not the same as Etherum. It's described in RSKIP-60, and it's compatible with EIP-55 ecnoding.

### Solution
- Implement RSK checksum in Ethereum `require_confirm_tx`
- Implement RSK checksum in `EthereumGetAddress`

### Implementation
- Line 20: SLIP-44 coin id derived from derivtion path (always positive)
- Line 37: lazy conditional avoids double checking `networks.by_`

### DIFF
--- a/src/apps/ethereum/get_address.py
+++ b/src/apps/ethereum/get_address.py
@@ -1,4 +1,5 @@
 from apps.wallet.get_address import _show_address, _show_qr
+from apps.ethereum import networks
 
 
 async def ethereum_get_address(ctx, msg):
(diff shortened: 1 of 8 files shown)


ilanolkies:
Isn't it related with https://github.com/trezor/trezor-core/issues/236?
`by_slip44` returns Undefined when 37310 is not found.

prusnak:
What's the purpose of `predefined_applying_rskip60`? Let's drop it.

Please add unit test for your code to

https://github.com/trezor/trezor-core/blob/master/tests/test_apps.ethereum.get_address.py

as `test_ethereum_address_hex_rskip60` method.

prusnak:
It's not required to close the pull request, just push (or push-force) into your branch and the pull request will get updated.

PR #248, opened 2018-06-20
embed/firmware: update linker to use firmware_extra section

prusnak:
These changes modify the linker script to utilize also the [firmware extra](https://github.com/trezor/trezor-core/blob/master/docs/memory.md) flash area as well. Currently, the MicroPython bytecode (`frozen_mpy.o`) is moved there (around 300K).

Old layout used almost full 768K in the firmware section and nothing was used in the firmware extra section (out of 896K) (i.e. 100% and 0%).

After the change the firmware section usage is around 470K and the firmware extra section usage is around 300K (i.e. 60% and  33%).

The resulting firmware is slightly bigger (because we need to pad the first section with zeroes) totalling at around 1070K (instead of the original 770K), but this provides space for future updates.

### DIFF
--- a/SConscript.firmware
+++ b/SConscript.firmware
@@ -298,6 +298,7 @@ env = Environment(ENV=os.environ, CFLAGS='%s -DPRODUCTION=%s -DPYOPT=%s' % (ARGU
 env.Tool('micropython')
 
 env.Replace(
+    CAT='cat',
     SED='sed',
     AS='arm-none-eabi-as',
     AR='arm-none-eabi-ar',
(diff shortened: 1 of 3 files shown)


SConscript.firmware
```diff
@@ -424,7 +425,9 @@ program_bin = env.Command(
     target='firmware.bin',
     source=program_elf,
     action=[
-        '$OBJCOPY -O binary -j .vendorheader -j .header -j .flash -j .data $SOURCE $TARGET',
+        '$OBJCOPY -O binary -j .vendorheader -j .header -j .flash -j .data --pad-to 0x08100000 $SOURCE ${TARGET}.p1',
```
prusnak: I don't like this `0x08100000` constant being here, but this was the most straightforward way how to do it.

Maybe there are some options to OBJCOPY which obsolete that and maybe also we can drop the CAT usage too.

jpochyla:
utACK

SConscript.firmware
jpochyla: I think it's OK for now.

PR #252, opened 2018-06-28
Add supporting Zencash

idimon4uk:
Hi, we have added support for ZenCash to trezor-core but we found that some of the existing unit tests are failing.

We made some investigation and found that a merge in trezor-common, commit: 6c540dbfff82c0cfcd23b09fe79ef6292e124ae1, moved values from 'bitcore' to 'blockbook' for some coins and these changes lead to breakdown of tests. The tests for our changes are completed successfully (we had to skip the other coins failing tests).


prusnak:
Please rebase the changes on top of the new master (changing your branch and force-pushing is OK, so this PR will get updated).

I renamed the field `coin.replay_protection_bip115` to `coin.bip115` so apply the change as well.

idimon4uk:
Hello Pavol,
We have removed the related issues and made code changes according to your comments.
Could you please review the updated pull request?

prusnak:
Thank you!

src/apps/wallet/sign_tx/scripts.py
```diff
@@ -42,6 +42,16 @@ def output_script_p2sh(scripthash: bytes) -> bytearray:
     return s
 
 
+def script_replay_protection_bip115(block_hash: bytes, block_height: bytes) -> bytearray:
+    block_height_size = len(block_height)  # SIZE in bytes block heigh (should be 3)
+    s = bytearray(38)
+    s[0] = 0x20  # 32 bytes for block hash
+    s[1:33] = block_hash  # block hash
+    s[33] = 0x03  # 3 bytes for block height
+    s[34:37] = (block_height if (block_height_size == 3) else block_height[:3 - block_height_size])  # MUST BE ONLY 3 BYTES FOR BLOCKHEIGHT
```
prusnak: Is this behaviour somewhere described? This does not seem very clever

idimon4uk:
Pavol, thank you for your comment. 

Replay protection needs a valid block height and a valid block hash, otherwise the transaction will not be spendable. If we rely on the correctness of the passed block height size, it is also possible to remove the check.
Regarding the following commit: (https://github.com/trezor/trezor-core/commit/9f90ca21a97bbb4d9f0629ed73cce4e83f50cbd9#diff-711e9d4abbb3c6798636eb0324377c3eL279) Please consider that Replay protection has been introduced in ZenCash after a certain block height, therefore not every transaction will have it specified in the script. For this reason, if someone tries to spend an old transaction, it will not be needed to consider replay protection block hash and block height to reconstruct the script to be signed. In this case it will be needed to keep the check. 

For example, please take a look at the following transaction: https://explorer.zensystem.io/insight-api-zen/tx/5dbcdfd462a6fef34e97ecc0d137bd172ad0be213ba985c1327af3ae3277ab95.

src/apps/wallet/sign_tx/scripts.py
idimon4uk: Pavol, thank you for your comment.

Replay protection needs a valid block height and a valid block hash, otherwise the transaction will not be spendable. If we rely on the correctness of the passed block height size, it is also possible to remove the check.
Regarding the following commit: (9f90ca2#diff-711e9d4abbb3c6798636eb0324377c3eL279) Please consider that Replay protection has been introduced in ZenCash after a certain block height, therefore not every transaction will have it specified in the script. For this reason, if someone tries to spend an old transaction, it will not be needed to consider replay protection block hash and block height to reconstruct the script to be signed. In this case it will be needed to keep the check.

For example, please take a look at the following transaction: https://explorer.zensystem.io/insight-api-zen/tx/5dbcdfd462a6fef34e97ecc0d137bd172ad0be213ba985c1327af3ae3277ab95.

prusnak: I get it. I fixed this behaviour in 6d5a8645fd1d1d4768d8b74b90125fd90b97bf79 - when None is passed to the script_bip115 function, an empty script is returned which is the behaviour you had originally.

Can you describe the magic about the `block_height` - where is the fact that it's always supposed to be 3 bytes documented?

idimon4uk: Pavol, it was a way to prevent to write a wrong block height if it was passed with more than 3 bytes. as explained before, if we rely on correctness of such block height size, we can remove the check.

prusnak: I think we should pass this value as `uint32` and convert to varint bytes, if you agree I'll do the change.

prusnak:
I changed `block_height` in bip115 to uint32 and corrected the processing of `block_height` param in `script_replay_protection_bip115` - commit 622eb001a68f4eb6c8c1150c1aeb3dbdaea7d992

src/apps/wallet/sign_tx/signing.py
```diff
@@ -275,8 +275,9 @@ class SigningError(ValueError):
                             multisig_get_pubkeys(txi_sign.multisig),
                             txi_sign.multisig.m)
                     elif txi_sign.script_type == InputScriptType.SPENDADDRESS:
-                        txi_sign.script_sig = output_script_p2pkh(
-                            ecdsa_hash_pubkey(key_sign_pub))
+                        txi_sign.script_sig = output_script_p2pkh(ecdsa_hash_pubkey(key_sign_pub))
+                        if(coin.bip115 and txi_sign.prev_block_hash_bip115 is not None):
+                            txi_sign.script_sig += script_replay_protection_bip115(txi_sign.prev_block_hash_bip115, txi_sign.prev_block_height_bip115)
```
matejcik: the fields `prev_block_hash_bip115` and `prev_block_height_bip115` are extracted from previous output's `scriptPubKey` apparently -- basically reversing what `script_replay_protection_bip115` does.

esp. given that this coin is not segwit-based and the prev output was actually requested somewhere near here (in `check_tx_fee`), ISTM the `prev_*_bip115` fields should be removed and Trezor should instead extract them itself

PR #255, opened 2018-06-30
Add BTG testnet config

wilsonmeier:


prusnak:
Address types collide with BTC values, please change them .

h4x3rotab:
We originally keep the testnet address value by intention because then we won't need to change a lot of testing stuff. It usually doesn't cause problem as it's just testnet. For bech32 address we have `tbtg`.

prusnak:
Please send a pull request to https://github.com/trezor/trezor-common/tree/master/defs/coins

That's how you add new coins to TREZOR.

PR #257, opened 2018-07-01
Add Groestlcoin support

yura-pakhuchiy:
Tests are not yet updated. Please review are these changes acceptable.


prusnak:
Some tests are failing in Travis, see here:
https://travis-ci.org/trezor/trezor-core/jobs/398907671
Most probably you just forgot to add API changes to the tests.

Also please add Groestl512 tests similar to
https://github.com/trezor/trezor-core/blob/master/tests/test_trezor.crypto.hashlib.sha512.py

yura-pakhuchiy:
@prusnak Thank you! I've fixed existing tests and  added some test vectors from groestl reference implementation: www.groestl.info/Groestl.zip

yura-pakhuchiy:
I've added tests for Groestlcoin addresses and transaction signing to ensure that changes to other coins will not break GRS accidentally.

prusnak:
Thanks! Looks great!

PR #258, opened 2018-07-02
Mutable Text API

jpochyla:


src/trezor/ui/text.py
```diff
@@ -102,17 +102,30 @@ class Text(ui.LazyWidget):
...
                  icon_color: int = ui.ORANGE_ICON):
         self.header_text = header_text
         self.header_icon = header_icon
-        self.content = content
+        self.content = list(content)
         self.new_lines = new_lines
         self.max_lines = max_lines
         self.icon_color = icon_color
 
+    def type(self, *content):
```
prusnak: Maybe `normal` would be better choice than `type` considering this is what it does right now?

jpochyla: 👍

PR #261, opened 2018-07-02
use pb2py from trezor-common

matejcik:
plus:
* WAY simpler `build_protobuf`
* common `pb2py` for our repos (also reusable for other kinds of protobufs)

changes in messages are just churn:
* fmt: off
* adding trailing commas on the argument list (this didn't work originally with `**kwargs` and I mistakenly kept it even when we threw kwargs away)
* `# type: ignore` on `List = None`
* different "default=" comments because we get them as strings now


prusnak:
Great!

PR #263, opened 2018-07-03
Automatic code style enforcement

jpochyla:
Uses black for source code formatting and isort for sorting imports. `make style` was modified to check for both.


prusnak:
utACK, nice!

PR #264, opened 2018-07-04
openocd: separate stlink configs deprecated - now combined

mcudev:
I noticed while running the latest commit of openocd (4999c9d98013eeb9746082f6b018e4b763ae3d37) the line:
`WARNING: interface/stlink-v2-1.cfg is deprecated, please switch to interface/stlink.cfg`

That file supposedly handles `STMicroelectronics ST-LINK/V1, ST-LINK/V2, ST-LINK/V2-1` now.
The commit message for that file says: `Extend HLA interface to allow multiple VID/PID pairs and use it to autodetect the connected stlink version.`

It seems to work with my dev kit (0483:374b STMicroelectronics ST-LINK/V2.1) and demo board (0483:3748 STMicroelectronics ST-LINK/V2) for erasing flash, re-flashing, and debugging - basically everything.

If it works for the openocd version that other developers are using please merge this. Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,7 @@ CROSS_PORT_OPTS ?=
 
 PRODUCTION ?= 0
 
-STLINK_VER ?= v2
-OPENOCD = openocd -f interface/stlink-$(STLINK_VER).cfg -c "transport select hla_swd" -f target/stm32f4x.cfg
+OPENOCD = openocd -f interface/stlink.cfg -c "transport select hla_swd" -f target/stm32f4x.cfg
 
 BOARDLOADER_START   = 0x08000000
 BOOTLOADER_START    = 0x08020000
--- a/docs/build.md
+++ b/docs/build.md
@@ -95,6 +95,5 @@ Use `make upload` to upload the firmware to a production device (with a bootload
 
 ### Flashing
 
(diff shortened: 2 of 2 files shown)


mcudev:
tangentially related to this, i added a faq, pinout, and photos for how to connect the debugger under reset to https://mcudev.github.io/trezor-model-t-dev-kit/index.html (search for "under reset" in that page).

i was only able to get "connect under reset" working with the ST-LINK/V2-B that was on my dev board. it might be because the cheapo separate ST-LINK/V2 usb debug adapter is not fully featured. but if this does not work with your V2 debugger, it might be a reason for you to upgrade debug adapters. with the dev kit used like this, you also get an external reset button.

or, this might not be too interesting :) connecting under reset is mainly useful in certain situations.

prusnak:
> connecting under reset is mainly useful in certain situations.

During my development I had several "bricked" boards which I was able to recover using "connect under reset", so this is definitively interesting.

I am aware of the fact that cheap ST-LINK clones don't provide this method and I had to use my original ST-LINK/V2.

prusnak:
Superseded by https://github.com/trezor/trezor-firmware/pull/45

PR #266, opened 2018-07-05
firmware: update flash_firmware target to use two-part flashing fix

mcudev:
binctl changes firmware.bin after it's put together from p1 and p2. so just need to copy back out p1 from firmware.bin.


prusnak:
Right! Thanks!

PR #267, opened 2018-07-06
build: make it easier to build binaries optimized for debugging

mcudev:
this is just a little development aid so that i don't have to keep patching the scons files to build binaries that are not optimized for size. the size optimized binaries are a pain to debug because the optimizations cause a lot of jumping around in the source.

example dev kit usage with this patch applied:
`pipenv run make clean vendor build_boardloader build_bootloader build_firmware flash openocd_reset DISPLAY_ILI9341V=1 STLINK_VER=v2-1 OPTIMIZE=-Og`

simply leave off the OPTIMIZE variable to get default current behavior.

### DIFF
--- a/SConscript.boardloader
+++ b/SConscript.boardloader
@@ -99,7 +99,7 @@ env.Replace(
     OBJCOPY='arm-none-eabi-objcopy', )
 
 env.Replace(
-    CCFLAGS='-Os '
+    CCFLAGS=env.get('ENV').get('OPTIMIZE', '-Os') + ' '
     '-g3 '
     '-nostdlib '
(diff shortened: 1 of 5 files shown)


matejcik:
utACK, this doesn't seem to affect anything unless `OPTIMIZE` is specified

mcudev:
yep. i just had an idea on how to make this even simpler. i'll have it pushed up a short while.

mcudev:
@matejcik just pushed up a simplified version that changes less but has the same usage and effects. if you like this, then we could do the same for the PRODUCTION flag too (not that it matters much besides making things a little easier to read/understand).

prusnak:
Thanks!

Issue #259, opened 2018-07-02
Implement Ripple


tsusanka:
We can get inspired by https://github.com/miracle2k/ripple-python

tsusanka:
You can track progress in the tsusanka/ripple branch

PR #272, opened 2018-07-11
Ripple support

tsusanka:
This introduces a support for Ripple. Only simple payments are supported. The only thing missing is a README, which I'll add soon.

- [x] add README
- [x] test on main net


tsusanka:
closes https://github.com/trezor/trezor-core/issues/259

PR #275, opened 2018-07-12
build: update to use latest arm none eabi toolchain gcc-arm-none-eabi-7-2018-q2-update

mcudev:
There is an updated gcc-arm-none-eabi-7-2018-q2-update toolchain available at https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads. They fixed the GDB TUI so it works again in this release (aids working in the debugger).

Things seem to work for building and debugging without any grief. I didn't see any extra built time warning messages either. I did NOT diff the binaries before and after (might later). Here's what I have verified works:

tested dev kit:
- `pipenv run make clean vendor build_boardloader build_bootloader build_firmware flash openocd_reset DISPLAY_ILI9341V=1 STLINK_VER=v2-1 OPTIMIZE=-Og`
- `make openocd STLINK_VER=v2-1`
- `make gdb_boardloader`

tested prototype hardware:
- `pipenv run make clean vendor build_boardloader build_bootloader build_firmware flash openocd_reset OPTIMIZE=-Og`
- `make openocd`
- `make gdb_boardloader`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
   global:
     - PROTOBUF_VERSION=3.4.0
   matrix:
-    - GOAL=stm32 TOOLCHAIN_SHORTVER=7-2017q4 TOOLCHAIN_LONGVER=gcc-arm-none-eabi-7-2017-q4-major
+    - GOAL=stm32 TOOLCHAIN_SHORTVER=7-2018q2 TOOLCHAIN_LONGVER=gcc-arm-none-eabi-7-2018-q2-update
     - GOAL=unix
     - GOAL=src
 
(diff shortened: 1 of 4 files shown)


mcudev:
I compared the bin output of the toolchains for the command: `pipenv run make clean vendor build_boardloader build_bootloader build_firmware`

build/boardloader/boardloader.bin, build/bootloader/bootloader.bin, and build/firmware/firmware.bin all had the exact same file size.

the boardloader bin files had matching sha256sum's (which is nice because the boardloader is the most important to not screw up).

the bootloader and firmware bin files were different between the versions of the toolchain. i eyeball compared `xxd` output and it's probably mostly address differences. i don't think it's anything to worry about.

prusnak:
Thank you for the extensive testing!

Issue #246, opened 2018-06-19
Not Flashing Properly on OSX


Rabitzwin:
I know I'm probably just missing a simple step somewhere, but I followed the hardware build guide and my "dev model" trezor (built with the microcontroller on the fancy testing board, the capacitive touch module/screen, and a micro sd reader with lots of soft wires) is not functioning. As far as I can tell I'm recieving proper feedback from the flashing command (included below), yet nothing appears on the screen and the trezor commands that I've been attempting to use from the library do not recognize this as a trezor device.

Any idea what could be wrong? Any and all feedback is appreciated (I apologize if this is the wrong way to ask for help, I didn't know how else to do it)

> openocd -f interface/stlink-v2-1.cfg -c "transport select hla_swd" -f target/stm32f4x.cfg -c "init; reset halt; flash write_image erase build/boardloader/boardloader.bin 0x08000000; exit"
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
hla_swd
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
adapter speed: 2000 kHz
adapter_nsrst_delay: 100
none separate
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
Info : clock speed 1800 kHz
Info : STLINK v2 JTAG v25 API v2 SWIM v14 VID 0x0483 PID 0x374B
Info : using stlink api v2
Info : Target voltage: 2.875485
Info : stm32f4x.cpu: hardware has 6 breakpoints, 4 watchpoints
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
adapter speed: 1800 kHz
target halted due to debug-request, current mode: Thread
xPSR: 0x01000000 pc: 0x08004b84 msp: 0x10010000
auto erase enabled
Info : device id = 0x20016419
Info : flash size = 2048kbytes
Info : Dual Bank 2048 kiB STM32F42x/43x/469/479 found
target halted due to breakpoint, current mode: Thread
xPSR: 0x61000000 pc: 0x20000046 msp: 0x10010000
wrote 49152 bytes from file build/boardloader/boardloader.bin in 1.698892s (28.254 KiB/s)
[...]

PR #278, opened 2018-07-14
docs: hardware: add photos, add more pin info, other related updates

mcudev:
here are some hardware doc updates that add photos and info for the new power pins and some other old, but unused things like the usb sbu pins.

i also verified that the sdram chip and the touch panel don't have to be removed from the STM32F429I-DISC1 dev board for things to work (although people may prefer to still remove those things and that is fine too).

i wired a couple of new boards from these instructions. they work. so we can probably close https://github.com/trezor/trezor-core/issues/246


prusnak:
Sweet!

Issue #183, opened 2018-04-04
update code in trezorhal for the new circuitry


prusnak:
- [x] 1. switch on/off SD card circuitry
- [x] 2. set SD card pins (DAT3:0, CLK, CMD) to log 0 or high Z when SD card is switched off
- [x] 3. switch on/off CTP (capacitive touch) circuitry
- [ ] 4. set CTP I2C pins (SDA, SCL) to log 0, others to high Z, when CTP is switched off

mcudev:
looks related to https://github.com/trezor/trezor-core/issues/86? i don't have great availability right now. but feel free to ping me if you need a second set of eyes or code review.

prusnak:
Yes, it is related. Will ping you once it's implemented. Thanks!

prusnak:
Pushed updated code to branch `new_hw`: https://github.com/trezor/trezor-core/tree/new_hw 

Applied points 1, 2 and 3. I don't do 4, because I start CPT immediately in touch_init.

mcudev:
Thanks for the update! Looks pretty good!

Observations:
 * Looks like you have pull-ups because the GPIOs default to being floating inputs. Seems correct.
 * Looks like we're then driving some P-channel enhancement mode MOSFET switches (active low).
 * Looks like more capacitance for the microSD power (would need to test and look at some datasheets to verify). Probably better.
 * Looks like the CTP is now not powered during reset and all the way through the boardloader and into the bootloader. I like that a lot.
 * Do we need a low-ohm series resistor between `TOUCH_ON` and the gate? It might not be necessary, I forget.

One thing I need to verify, but am pretty sure that I did on purpose for the display code after scoping it -- see this example where I make sure to write the pin state before I call `HAL_GPIO_Init` https://github.com/trezor/trezor-core/blob/new_hw/embed/extmod/modtrezorui/display-stm32.h#L170-L171

For `SD_ON`, we'd want to make sure that is `GPIO_PIN_SET` before init-ing it as an output. Now we might be ok if the default is to output a high. Easy enough too verify (maybe tomorrow).

Same feedback for `CTP_ON`. I'd probably prefer `HAL_GPIO_WritePin` before `HAL_GPIO_Init`.

Also, now that there is a switch and we can control the whole startup sequence of the CTP, there may be some more that we can do with that better control. I get the feeling that I had something in the back of my mind when I called adding the switch a potential upgrade. Would love some new dev hardware for testing, if you've got some extra prototypes :)

[...]

PR #279, opened 2018-07-15
boardloader: partial fix for powering off sd card

mcudev:
Referencing https://github.com/trezor/trezor-core/issues/183 and https://github.com/trezor/trezor-core/issues/86

I was testing with the new prototype board and found this bug.
Running build of current master: `pipenv run make clean vendor build_boardloader build_bootloader build_firmware flash openocd_reset OPTIMIZE=-Og`

I found this bug by starting up with the sd card inserted.
During 10 second countdown eject sd the card (or rapidly eject and re-insert -- it's a race condition).

You'll get a red screen fatal error, like:
```
expr: sdcard_power_on()
file: embed/boardloader/main.c:52
func: check_sdcard
rev: 95416629
```

Re-insert the sd card and the system resets.

I think this bug shows that the problem of in-rush current causing a power fault while inserting a micro sd card while the sd power is on is still not entirely solved.

Maybe the circuitry should only allow the MOSFET to turn on if the gate is low AND the SD card detect (debounced) is low?

To simulate the race in the debugger:
- plug in debug adapter
- plug in prototype board
- make openocd
- make gdb_boardloader
- insert bootloader formatted micro sdcard
```
(gdb) b *copy_sdcard
(gdb) c
(gdb) b *sdcard_power_on
(gdb) c
(gdb) b *sdcard_is_present
(gdb) c
(gdb) lay src
(gdb) bt (note that the micro sdcard is powered on at this point)
#0  sdcard_is_present () at embed/trezorhal/sdcard.c:108
#1  0x08009162 in sdcard_power_on () at embed/trezorhal/sdcard.c:117
#2  0x08005c04 in check_sdcard () at embed/boardloader/main.c:52
#3  0x08005d0e in copy_sdcard () at embed/boardloader/main.c:97
#4  0x08005ece in main () at embed/boardloader/main.c:187
```
- eject sd card
- `(gdb) c`
- get the fatal error screen
- insert card and get system reset

This PR fixes this particular bug. With it, the system will not reset because the SD power switch is turned off. But, it does not fix for other similar cases, like ejecting the sd card in the middle of the boardloader's copying of the bootloader from the sdcard like at line `ensure(sdcard_read_blocks(buf, i, 1), NULL);`
That's why I wonder if the circuitry needs to be AND'ed with the card detect circuit.

### DIFF
--- a/embed/trezorhal/sdcard.c
+++ b/embed/trezorhal/sdcard.c
@@ -115,7 +115,7 @@ secbool sdcard_power_on(void) {
     HAL_Delay(50);
 
     if (sectrue != sdcard_is_present()) {
-        return secfalse;
+        goto error;
     }
     if (sd_handle.Instance) {
         return sectrue;
@@ -147,17 +147,15 @@ secbool sdcard_power_on(void) {
     return sectrue;
 
 error:
-    sd_handle.Instance = NULL;
+    sdcard_power_off();
     return secfalse;
 }
 
 void sdcard_power_off(void) {
-    if (NULL == sd_handle.Instance) {
-        return;
+    if (sd_handle.Instance != NULL) {
+        HAL_SD_DeInit(&sd_handle);
+        sd_handle.Instance = NULL;
     }
-    HAL_SD_DeInit(&sd_handle);
-    sd_handle.Instance = NULL;
-
     // turn off SD card circuitry
     HAL_Delay(50);
     sdcard_default_pin_state();


mcudev:
This is worse than I realized. In production, if my prototype board works like the rest, it will usually fail `reset_flags_check` and black screen when the sd card is present. Basically, switching the SD MOSFET on is causing a power fault still.

Put a breakpoint at `reset_flags_check` with: 
`(gdb) b *reset_flags_check`
`(gdb) display /x RCC->CSR`
sometimes I get `PINRSTF` and sometimes I get `PINRSTF | BORRSTF` when `PC0` is set low in `sdcard_power_on`. For example:
```
1: /x RCC->CSR = 0x0
(gdb) bt
#0  sdcard_power_on () at embed/trezorhal/sdcard.c:114
#1  0x08005c04 in check_sdcard () at embed/boardloader/main.c:52
#2  0x08005ec8 in main () at embed/boardloader/main.c:186
(gdb) n
halted: PC: 0x08009190
halted: PC: 0x08009192
halted: PC: 0x08009194
halted: PC: 0x08006828

Breakpoint 4, reset_flags_check () at embed/boardloader/lowlevel.c:165
1: /x RCC->CSR = 0x4000000
```

mcudev:
From https://resources.altium.com/pcb-design-blog/how-to-design-microsd-power-circuits-without-destabilizing-on-board-voltage-supply, some things to try:
- "[capacitance] values above 45uF will do"
- "use an FDN340P MOSFET to control the power supply of the microSD" which I suppose is a MOSFET that they've tested with an acceptable voltage slew rate
- check into something like this? http://www.mouser.com/ds/2/149/FDG901D-94072.pdf

prusnak:
Merged the fix, but I am aware of the outstanding issues and I'll report my further investigations.

prusnak:
I tried this with 3 boards I have on my desk and have not seen the behavior you are describing ...

My test:

1) checkout master (fecc6e53a0ca45a852d9964c6ba1a3a0a35ae4e9)

2) comment lines https://github.com/trezor/trezor-core/blob/master/embed/boardloader/lowlevel.c#L156 and https://github.com/trezor/trezor-core/blob/master/embed/boardloader/lowlevel.c#L163 (so we check `RCC->CSR` in non-PRODUCTION build)

3) `make build` and `make combine` and `make flash_combine`

Then I run the device without SD card (boots to firmare) and with SD card (boardloader shows the countdown).

Am I missing something? Maybe this depends on SD card used?

mcudev:
I can still reproduce with `fecc6e53a0ca45a852d9964c6ba1a3a0a35ae4e9`.

It could potentially be the microSD card I'm using. Maybe it has less favorable electrical characteristics. I'll try another in a bit. The one I am using here is a Kingston 32GB microSD HC model SDCA3/32GB (https://www.amazon.com/Kingston-Digital-32GB-microSDHC-SDCA3/dp/B00RVNGGT8).

- I reset my branch to be exactly fecc6e53a0ca45a852d9964c6ba1a3a0a35ae4e9
- I commented out the two lines you listed
- I built and flashed it to my board (`pipenv run make clean vendor build_boardloader build_bootloader build_firmware flash openocd_reset OPTIMIZE=-Og`)
- I unplugged the debugger and the main USB cable
- I made sure that the microSD card was removed and plugged in main USB -- things work as normal
- I unplugged main USB, inserted microSD, plugged main USB and screen stays black

I did this a few times and it was reproducible.

Try this:
- reset your branch to be exactly fecc6e53a0ca45a852d9964c6ba1a3a0a35ae4e9
- build and flash
- insert microSD card
- `make openocd`
- in a new window: `make gdb_boardloader`
```
(gdb) b *reset_flags_check
(gdb) c
(gdb) display /x RCC->CSR
(gdb) c
```

If the breakpoint hits a second time, you know that you went through another reset and you'll get the RCC->CSR output to see the circumstances of the reset.

Here's what I see on my run that shows a brownout reset having happened when you see at the end `RCC->CSR = 0x6000000`:
```
(gdb) b *reset_flags_check
Breakpoint 1 at 0x8005bc0: file embed/boardloader/lowlevel.c, line 165.
(gdb) c
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.

Breakpoint 1, reset_flags_check () at embed/boardloader/lowlevel.c:165
165	    RCC->CSR |= RCC_CSR_RMVF; // clear the reset flags
(gdb) display /x RCC->CSR
1: /x RCC->CSR = 0x14000000
(gdb) c
Continuing.
halted: PC: 0x08005bc2

Breakpoint 1, reset_flags_check () at embed/boardloader/lowlevel.c:165
165	    RCC->CSR |= RCC_CSR_RMVF; // clear the reset flags
1: /x RCC->CSR = 0x6000000
```

mcudev:
NOT reproducible with Samsung 182 EVO+ Micro SDXC MB-MC128D (https://www.amazon.com/Samsung-128GB-UHS-i-Adapter-MB-MC128D/dp/B00WR0HIR4). So, I guess the Kingston card might draw more current? I'll try some more cards...

mcudev:
Could NOT reproduce with a SanDisk 32GB microSD HC UHS-I U1 Class 10.

Could NOT reproduce with a SanDisk Extreme 32GB microSD XC UHS-I U3 sdsqxne-032g-gn6ma (https://www.amazon.com/SanDisk-Extreme-microSDXC-Adapter-SDSQXNE-032G-GN6MA/dp/B013CP5HCK).

mcudev:
BTW, thanks for merging and testing the partial fix!

The Kingston card seems to work for me in other places that I use it, so I can't just write it off as being a bard card. Maybe you want to get one and try it? If it fails also, then you have a good card for testing closer to the limits (again, assuming that model operates within the microSD spec limits).

In summary, it appears that the microSD subsystem still causes power faults for some subset of cards. The two potential ways that I can think of to fix that are 1) to add more capacitance to the power rail near the microSD and 2) to reduce the slew rate.

There are still some cases in the code where the microSD power switch is left on even after the card has been removed. There is no circuit to automatically turn it off and we use polling instead of an interrupt to handle the card detect + power on/off. I'm kind of not very excited to try to handle it with an interrupt due to debouncing and trying to keep the boardloader as simple as possible to follow.

I'm not sure how far along production with this hardware is, so let me know what you're thinking. If hardware changes are still possible, then I'll also add that maybe a switch for the display power would be neat too. Thanks!

prusnak:
Was able to find a card that is also behaving oddly. Boardloader with bootloader on SD card will start, show countdown - number 10, LDO Voltage Regulator goes really hot and afterwards the screen goes black (I suspect BOR reset).

The card is *Kingston Micro SDHC SDC4/16GB*. I have two of them and behave exactly the same.

OTOH Kingston SDC10G2/8GB works fine. Also I have tried numerous 1GB SD cards (not SDHC) by A-Data and they worked also fine.

I will give the problematic SD cards to our HW development lab, so they can investigate the issue.

> I'm kind of not very excited to try to handle it with an interrupt due to debouncing and trying to keep the boardloader as simple as possible to follow.

Same here, let's not complicate things more than needed.

> If hardware changes are still possible, then I'll also add that maybe a switch for the display power

I would avoid the changes for this production run. But we'll keep revisiting the design, so next production run might include this change too, but not this one.

mcudev:
Yep, agree with all that, thanks!

prusnak:
I refactored some stuff in sdcard/touch so these two are now more similar - see https://github.com/trezor/trezor-core/tree/touch_on_off. It also allows us to do `touch_init`, so we can properly set CTP pins in boardloader without turning it on. This is now done using `touch_power_on`/`off`.

What do you think?

prusnak:
Hm, it seems that the touch is working when I use `init`, then `power_on`, but when I call `power_off` and `power_on` again, it won't work again :-(

mcudev:
`i2c_handle` is a .bss section variable that gets initialized to zeros already, so I would remove the reference to `i2c_handle` by `touch_init` to clarify that it's only manipulating GPIOs.

My thoughts on calling `touch_init` from within the boardloader:
- it's more flexible to have CTPM code in the bootloader and later stage code where it's upgradeable and not write protected.
- since the GPIO's are floating inputs coming out of reset, the pull up should ensure that the CTPM is powered off during the boardloader already. so it'll be off with or without this in the boardloader.
- we currently don't use touch input or need it in the boardloader. this code change doesn't power on the ctpm, but i'd rather not open up the possibility of interacting with the ctpm from within the boardloader (with a future code change) by even just init'ing the GPIO's right now. i'm still wary of the external touch panel and display driver chips. their docs are not great, and i'm suspicious of the ctpm driver chip's potential to be used by someone craftier than myself. i think of it like we have to treat the FT6236 as any other untrustworthy external device. we can be careful with it in later stages, but i'd rather not deal with it in the boardloader.

i suppose the benefit is that we only have to call `touch_init` once in the boardloader and then never again.
but if we change the pins used in the future (e.g.- start using PC4), then the initialization gets spread out into later stage code anyways because of the write protection.

additional tangential thoughts:
- i feel the same way about the SDIO peripheral for the microSD, but i realize that it's a necessary evil to avoid a device being bricked when the bootloader ever gets corrupted. at least the sd card is removable and thus the bulk of the SDIO code does not run most of the time. plus it's anon-chip peripheral so i feel a little better about it.
- the primary purpose of init'ing the display in the boardloader is to clear the GRAM. of course we need the display available to show any boardloader error messages too. but the primary concern for me was wiping the display memory. i think of the boardloader as our secure little area to take care of things like that (modulo invasive/sophisticated physical attack). this is part of the reason why i keep asking about a display power switch and a default off display :) the tertiary concern is avoiding screen flicker and giving a snappy and user experience.

prusnak:
Hm, it got me thinking and did some changes:

1) removed the touch stuff from the boardloader again
2) removed `{sd_handle,i2c_handle}.Instance = NULL;` from `sdcard_init`/`touch_init`
3) move turning on SD card circuitry in `sdcard_power_on` after the `sdcard_is_present` check (we don't need to turn on the whole sd card circuit just to check the sd_present pin)

I still observe the problem the CTP is not working after the following sequence: init, on, (works), off, on (does not work). Investigating.

prusnak:
I am also thinking about moving the call to `{sdcard,touch}_default_pin_state` to the beginning of `{sdcard,touch}_init` function.

mcudev:
1. ack
2. ack
3. ack

Will need to add a `touch_power_on();` in  embed/firmware/main.c just after `touch_init();`

Regarding, `touch_default_pin_state` I don't think you need to drive the alternate function open drain pins PB6 & PB7. With that in mind, I think moving `touch_default_pin_state` to the beginning of `touch_init` is ok, then you can remove the inline `HAL_GPIO_WritePin` later in the function.

Similar feedback for `sdcard_default_pin_state`. I don't think we need to `HAL_GPIO_WritePin` for PC8-12 and PD2 since they are alternate function inputs.

> I still observe the problem the CTP is not working after the following sequence...

I have not looked into this yet, but I can.

prusnak:
> Will need to add a touch_power_on(); in embed/firmware/main.c just after touch_init();

Ack

> I have not looked into this yet, but I can.

Already fixed

> I don't think we need to HAL_GPIO_WritePin for PC8-12 and PD2 since they are alternate function inputs.

I was told that SD circuit gets still voltage from DATA pins after the SD_ON PIN is turned off. Maybe we should set these pins to OUTPUT in the init?

mcudev:
Do the `HAL_GPIO_WritePin` calls for those pins in `sdcard_default_pin_state` even have an effect or are they ignored because of being in alternate function mode?

prusnak:
I set pins to OUTPUT default_pin_state and set them to required values in new active_pin_state => https://github.com/trezor/trezor-core/commit/ced85b014f6ee6676f58640d65e7a4822c3bc919

This solved my issue with the Kingston 16GB card.

mcudev:
Interesting that that worked for you. I'll review this closer in a little bit.

I just put ced85b014f6ee6676f58640d65e7a4822c3bc919 onto my board and unplugged my board entirely. Insert microSD card and plug in the main USB cable (not the debug adapter). I get a black screen until I eject the microSD which I am able to reproduce reliably causes:

```
FATAL ERROR:
expr: sdcard_power_on()
file: embed/boardloader/main.c:52
func: check_sdcard
rev: ced85b01
```

I then verified in the debugger that I'm getting the brownout reset still with the `reset_flags_check` breakpoint and `RCC->CSR` display.

I set a break point on `sdcard_power_on` and when I step into `sdcard_active_pin_state`, attempting to run `HAL_GPIO_WritePin(GPIOC, GPIO_PIN_0,  GPIO_PIN_RESET);` resets my board.

This is all with my **Kingston 32GB microSD HC model SDCA3/32GB**.

prusnak:
The last thing on my mind is the cable. Maybe you are using a very long cable - I tried both 1 m A-DATA USB A-C cable and the same USB A-C cable we ship in the box (30 cm?). Dunno ...

mcudev:
The cable I'm using is about 26cm tip to tip. I just tried the cable that came with the production Model T and got the same results.

I also just noticed that this latest commit has `sdcard_is_present` being used before it's been configured with its `GPIO_PULLUP`. So, it's a floating input which may cause spurious results early in the boardloader.

prusnak:
> I also just noticed that this latest commit has sdcard_is_present being used before it's been configured with its GPIO_PULLUP. So, it's a floating input which may cause spurious results early in the boardloader.

Good catch. Fixed in baa6fef875f64277c0ea6e4d816aed72aebeef80

mcudev:
ok, i'm going to tinker with this for a while. i'll get back to you a bit later.

prusnak:
Thank you!

mcudev:
Do you get the same benefit for your Kingston card by only init'ing PC9 in `sdcard_default_pin_state`? Like this:
```
-    GPIO_InitStructure.Pin   = GPIO_PIN_0 | GPIO_PIN_8 | GPIO_PIN_9 | GPIO_PIN_10 | GPIO_PIN_11 | GPIO_PIN_12;
+    GPIO_InitStructure.Pin   = GPIO_PIN_9;
```

prusnak:
Do you get the same benefit for your Kingston card by only init'ing PC9 in sdcard_default_pin_state?

No, it still halts for the problematic card.

mcudev:
gah! there's an external pullup on PC9 on this dev board. nevermind that.

mcudev:
> I just got report from the lab that the changes fix the issue because SD_ON is set to log 0 (circuit active) most of the time...

I feel like I'm misreading that. Do you mean that the external pullup on PC0 is not effective and so driving it high in `sdcard_default_pin_state` is necessary?

prusnak:
This is what they see:

sd_on

SD_ON is log 0 most of the time (SD circuit enabled). It goes to log 1 just for a while, then it goes back.

It should be log 1 most of the time (SD circuit disabled) and go to log 0 when interaction with SD card is needed.

mcudev:
ok, so then i guess that the 25ms of being high is between the middle of  `sdcard_default_pin_state` and the beginning of `sdcard_active_pin_state`? that would seem to indicate that the external pullup is not keeping the gate to the MOSFET high, right?

prusnak:
Do you suggest to use PULLUP for PC0 output or changing the R2/R3 value or all of this?

mcudev:
I think the resistors external to the MCU because the PC0 pin will be a floating input for some amount of time before we could set its pullup. I was assuming that the external pullup was keeping the gate high already. So, maybe it turns out that it isn't strong enough of a pullup right now?

prusnak:
Yes, that's what I thought.

mcudev:
For reference, I verified with page 795 of UM1725 about the HAL SD Generic Driver that we are doing as the manual says to do (https://www.st.com/content/ccc/resource/technical/document/user_manual/2f/71/ba/b8/75/54/47/cf/DM00105879.pdf/files/DM00105879.pdf/jcr:content/translations/en.DM00105879.pdf):
"Configure these SDIO pins as alternate function pull-up"

mcudev:
If the circuit is working as I assumed it would, then the PC0 behaving as in the scope for the `#if 1` case would not have the SD circuit being powered almost all the time. The left part of the scope before the rise would be PC0 as a floating input and then SD circuit would be off (because of the external pullup) until PC0 is driven low in the scope output.

I'll see if I can get it on a scope with my debugger and step through. It'll be difficult because everything's so small and close together and I don't have a great toolset. Do you have any docs on the location of test points on the PCB (perhaps the brd file)?

mcudev:
I'll do more digging over the weekend.

This might be nothing, or it might be something, I don't know yet.
I was poking around with my scope and saw that pin 2 on the microSD socket (CD/DAT3) had different behavior. Section 5.1.1 in http://www.mouser.com/ds/2/302/AN10911-112956.pdf describes that pin.

The HAL code does not appear to disable the card detect pull-up internal to the sd cards. Maybe that matters for some of them?

I found some links with seemingly good related info:
https://electronics.stackexchange.com/questions/39571/how-to-do-pulling-up-or-down-correctly-when-interfacing-a-microsd-card
https://electronics.stackexchange.com/questions/224200/sd-card-acmd42-command
https://www.kingston.com/datasheets/SDCIT-specsheet-8gb-32gb_en.pdf
https://embdev.net/attachment/39390/TOSHIBA_SD_Card_Specification.pdf
https://github.com/LonelyWolf/stm32/blob/master/cube-usb-msc/sdcard-sdio.c

mcudev:
Figured that I'd share this. With my problem card, the **Kingston 32GB microSD HC model SDCA3/32GB**, this is the brownout that I see when probing the **source pin** of the microSD MOSFET when I run this line: https://github.com/trezor/trezor-core/blob/0f7a90dfe51236ae595e82a6896317d1dd4a0e84/embed/trezorhal/sdcard.c#L97
fault

mcudev:
I verified that the gate pin on the sdcard MOSFET is high until we set PC0 low.
I verified that the gate pin on the CTPM MOSFET is high until we set PB10 low.
The external pullups appear to be keeping the switches off while the pins on the MCU are still floating inputs (under reset and until we configure and use them to drive the gate pins).
That's all good. That's what I had assumed (and was hoping) was going to happen based on the schematic.
I also probed around TP15 and TP16 and the source and drain pins of the MOSFETs and all confirm those verifications.
I did this while stepping through with the debugger.

With your `#if 0` case and a card that works for me, the sdcard socket 3V3 pin measures about 1V and the AF pullup pins measure about 1.4V while PC0 is high (MOSFET is off).
When I run the same measurements with my problem card, the same pins measure about 400mV lower at the same points in the code.

I also did some tinkering with the ACMD42 card detect resistor. It's not the cause of our main issue here, but we might want to disable that resistor so that we don't run into any issues because of it. This [doc](http://envoydatamemory.com/datasheets/EN-L0J%20Industrial%20microSD%20Card%20Spec%20Rev1.6.pdf) said a little more about DAT3.

prusnak:
We (hopefully) fixed the issue with our card today, see https://github.com/trezor/trezor-core/commit/08725c16e857fc18e3221b8f331a1c8903caf7a6 (`hw` branch now merged into `master` and deleted).

SD DATA pins are set to OUTPUT/NOPULL/log0 when inactive and we added `HAL_Delay(10);` after turning on the SD card circuit via PC0. After these changes, the circuit behaves correctly (checked with an oscilloscope).

mcudev:
My problem card still acts the same way with these changes.

With my good card, I can cause a fatal error by ejecting it during the boardloader 10 second count down (doesn't happen every time, so probably a race like before):
```
FATAL ERROR:
expr: sdcard_power_on()
file: embed/boardloader/main.c:52
func: check_sdcard
rev: 8cf0b93d
```

prusnak:
Can you try adding 2,2nF capacitor between gate and drain of T1?

screenshot from 2018-07-23 16-03-00

mcudev:
I'm working on it. This is more of a dexterity test :) I need to go digging to find more caps. So, far I've tried a 22pF and a 10,000pF and every once in a while the board does not reset. It's a crude test because I'm literally pressing the ceramic cap leads onto the pins and reaching over to continue the debugger.

PR #288, opened 2018-07-24
Fix U2F compatibility with Trezor One

### LINKED ISSUE #276: U2F is incompatible between One and T
> recovered my seed on model T (previously used in model One) and website refused to accept this U2F device "another device was used to register".

Confirmed

> It is supposed to work that way, but I think it is not thoroughly tested. I just tried it and I get the same problem.

For reference https://www.reddit.com/r/TREZOR/comments/8y2hbr/u2f_and_model_t/

jpochyla:
Fixes #276 in a backwards-compatible manner.


prusnak:
utACK

PR #291, opened 2018-07-26
touch: follow ft6236 power on sequence closer

mcudev:
- removed the inline attribute on `touch_default_pin_state` and `touch_active_pin_state` to make setting debug breakpoints easier
- follow direction of section 3.5 of FT6236 datasheet regarding power on sequence
- setup PC4 and the interrupt polling mode (even though it's still not used). i did this because the datasheet recommended to set the pin low during power off. so, i also configured it for when power was on.
- moved the 100ms wait into `touch_default_pin_state` to account for when `touch_init` is called at the beginning of the firmware. that goes through an implicit power off with the setting of the default for the CTPM power pin.
- changed to keep the CTPM in reset state after power-on until all the other pins are configured

- i did try polling the interrupt line at the beginning of the bootloader to see if it would go low just by touching before plugging in and then holding your finger still (to get into the bootloader) and that still did not work. oh well. users still have to move their finger around to get into the bootloader.

- i tested this PR with my v12 and v10 prototype boards. it seemed to work with both. but now that i'm thinking about it, i wonder if the way that we refactored the touch and sdcard stuff to drive low outputs will have negative impact on the older boards. we have to account for the old devices always having the CTPM and sdcard powered. is that going to be a problem? how should we handle that? i'm too tired to think of it right now -- just leaving a note so it doesn't get forgotten.


mcudev:
> i wonder if the way that we refactored the touch and sdcard stuff to drive low outputs will have negative impact on the older boards

i did some testing and looked at the schematic differences and didn't notice any problems. i guess this is ready.

prusnak:
Checked and it's fine. Thanks!

PR #286, opened 2018-07-23
extmod: Monero module added (trezor-crypto binding) [for review]

ph4r05:
Extmod with trezor-crypto monero binding.
Depends on https://github.com/trezor/trezor-crypto/pull/162 

Provides python binding to the trezor-crypto monero related code.

PR #293, opened 2018-07-27
Monero support added [for review]

ph4r05:
The first  review attempt for Monero code.

Code can be tested with monero-agent as with the previous PoC:
https://github.com/ph4r05/monero-agent

Documentation for PoC (section "Testing with Trezor"):
https://github.com/ph4r05/monero-agent/blob/master/PoC.md

For testing you may actually need my forks as I use my dependencies with PRs merged.
https://github.com/ph4r05/trezor-core/tree/xmr-total-full

This PR depends on PRs:

https://github.com/trezor/trezor-crypto/pull/162
https://github.com/trezor/trezor-core/pull/286


src/apps/monero/xmr/key_image.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def compute_hash(rr):
+    """
+    Hash over output to ki-sync
+    :param rr:
+    :type rr: TransferDetails
+    :return:
+    """
+    buff = b""
```
ph4r05: Will optimize this

src/apps/monero/xmr/crypto.py
```diff
@@ -0,0 +1,663 @@
...
+# https://godoc.org/github.com/agl/ed25519/edwards25519
+# https://tools.ietf.org/html/draft-josefsson-eddsa-ed25519-00#section-4
+# https://github.com/monero-project/research-lab
+
+import ubinascii as binascii
+
+from trezor.crypto import hmac, monero as tcry, pbkdf2 as tpbkdf2, random
+from trezor.crypto.hashlib import sha3_256
+
+# py constants
```
ph4r05: Will be removed

tsusanka:
Hey @ph4r05, I'l

tsusanka:
Hey @ph4r05, I'll be doing some sort of preliminary review with @jpochyla doing review of my review later :). Consider it more as questions/suggestions. First, I would like to clarify few things:

- The https://github.com/ph4r05/trezor-core/tree/xmr-total-full branch includes this PR alongside with all dependencies merged, is that correct?
- What is the difference between https://github.com/ph4r05/trezor-core/tree/xmr-total-full and https://github.com/ph4r05/trezor-core/tree/xmr-total-2 ? Is the xmr-total-2 the same just with the commits squashed and no dependencies?
- If I would like to test it all, I can use the xmr-total-full branch and follow the instructions in monero-agent:PoC.md@master, correct?
- Is it okay if I merge this into a 'monero' branch in the very same manner as @prusnak has done with trezor-crypto? It feels bit hostile to replace your PR with mine, but it does make modifications easier.

ph4r05:
OK thanks!

- `xmr-total-full` branch is basically my development branch with lots of commits and with my custom git submodule forks of `trezor-common` and `trezor-crypto` with all required stuff merged in.

- `xmr-total-2` is all monero work done in one squashed commit, rebased on top of the master. It has original git submodules so this branch tests are failing, e.g. there is no monero functionality in `trezor-crypto`.

- Regarding the testing - yes. If you fork my `xmr-total-full` with all my submodules and follow PoC.md@master you should be able to both run trezor-specific tests (16 testing transactions) and sign transactions with agent as you did in the PoC phase of the project.

- PR merge - if the proposed way is easier for you then I think its OK. I am not sure how github / @prusnak is doing it in https://github.com/trezor/trezor-crypto/pull/169 but there is also my avatar alongside the modified commits so its fine :)

ph4r05:
BTW 

- is there a way to set a timer after reception of a message, e.g. for a minute so if no next protocol message is received within the timeout the whole state is wiped out?

- what is the safest way to wipe out secrets from the memory allocated in the python code? E.g., spending keys for UTXO would be nice to zero out before garbage collecting. Is the currently only way to do it manually?

tsusanka:
@ph4r05 there is no such file `trezor-crypto/monero/crypto.c` as required in SConscript.firmware. This is probably due to https://github.com/trezor/trezor-crypto/pull/169#issuecomment-409237456. I've modified the line to `trezor-crypto/ed25519-donna/ge25519.c`, however this still fails due to `"ROTL32" redefined` error. Would you like to look into it?

ph4r05:
Hmm I think this PR is not directly usable with `trezor/trezor-crypto` until all changes are merged. As they are still ongoing changes and review process I decided not to keep it up-to-date with the `trezor/trezor-crypto` until the merging and review is finished.

For the testing purposes, you may use the branch `ph4r05/xmr-total-full` with working `trezor-crypto` changes. 

That should also fix the `"ROTL32" redefined` error as it is defined in `ed25510-donna.h` which did not have include-once ifndef

tsusanka:
Ok fair enough. Let's wait then. @prusnak could you answer @ph4r05's questions above?

tsusanka:
note: the monero branch includes this PR and is rebased on top of current master

prusnak:
> is there a way to set a timer after reception of a message, e.g. for a minute so if no next protocol message is received within the timeout the whole state is wiped out?

No, this is not yet possible.


> what is the safest way to wipe out secrets from the memory allocated in the python code? E.g., spending keys for UTXO would be nice to zero out before garbage collecting. Is the currently only way to do it manually?

Your request sparked an idea, which I implemented here: https://github.com/trezor/trezor-core/commit/ebf912c8f1af2557778ece973f3e471d287f303c

Since our modtrezorcrypto objects implement memzero descructors in `__del__`, we can use the following syntax to acheive what you want:

```
from trezor.crypto import SecureContext

with SecureContext() as sc:
    sc.var1 = ...
    sc.var2 = ...
```

Later we might move this variables to different parts of the CPU memory (CCRAM instead of main RAM) or something even more sofisticated.

ph4r05:
I've refactored transaction signing so it uses protob serialization where suitable - to avoid cryptonote serialization for complex messages which are not forward compatible.

ph4r05:
From now I will stop squashing commits after each change. I will squash it to one eventually after the review.

ph4r05:
@tsusanka Change address is not confirmed by user now, it is verified to match primary account address (i.e., subaddress with index (account, 0)).

ph4r05:
@tsusanka I've added tests for Monero Cryptonote serializer. For that I needed to add support for async test cases, which is also in a separate PR:

https://github.com/trezor/trezor-core/pull/319

ph4r05:
I've changed trezor-crypto to my fork until the required changes are merged into trezor-crypto, namely:

- https://github.com/trezor/trezor-crypto/pull/173
- https://github.com/trezor/trezor-crypto/pull/169
- https://github.com/trezor/trezor-crypto/pull/174 OR https://github.com/trezor/trezor-crypto/pull/175

The commit of the trezor-crypto submodule has marged all required PRs so this PR can be of any use / tested.

ph4r05:
btw @prusnak another thing I found very useful w.r.t. memory saving is to wrap certain blocks with:

```python
from trezor import utils

mods = utils.unimport_begin()
result = memory_heavy_computation()
utils.unimport_end(mods)
```

As this sweeps also some packages imported in the `memory_heavy_computation()`. It would be maybe also nice to have some context manager for that, like:

```python
with utils.mods_block():
    result = memory_heavy_computation()
```

prusnak:
@jpochyla What do you think about the comment above? Context manager for
memory heavy stuff seems like a nice idea

ph4r05:
Adapted to master for trezor-crypto.
Travis fails because of some problem on trezor-common master.

```
The command "test "$GOAL" != "src" || pipenv run make templates_check" exited with 2.
```

tsusanka:
The tests should be fixed now.

src/apps/monero/diag.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Author: Dusan Klinec, ph4r05, 2018
+
+import gc
+import micropython
+import sys
+
+from trezor import log
+
```
jpochyla: Same here, can we either get rid of this module or wrap it in `__debug__`? Is is useful to have it in production?

src/apps/monero/protocol/key_image_sync.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Author: Dusan Klinec, ph4r05, 2018
+
+from trezor import log
+
+from apps.monero.controller import wrapper as twrap
+
+
+class KeyImageSync(object):
```
jpochyla: `class KeyImageSync:` would be best here.

src/apps/monero/sign_tx.py
```diff
@@ -0,0 +1,43 @@
...
+
+    from apps.monero.protocol.tsx_sign import TsxSigner
+
+    log.debug(__name__, "TsxSigner. F: {} A: {}".format(gc.mem_free(), gc.mem_alloc()))
+    log.debug(__name__, "TsxState: %s", state.ctx_sign)
+    gc.collect()
+
+    try:
+        signer = TsxSigner()
+        await signer.wake_up(ctx, state.ctx_sign, msg)
```
jpochyla: Both `TsxSigner.__init__` and `TxSigner.wake_up` create new `TrezorInterface`.

src/apps/monero/xmr/common.py
```diff
@@ -0,0 +1,108 @@
...
+# -*- coding: utf-8 -*-
+# Author: Dusan Klinec, ph4r05, 2018
+
+
+from trezor.crypto import monero, random
+
+
+class XmrException(Exception):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
jpochyla: I think we can omit this constructor (for `pass`).

src/apps/monero/xmr/common.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def random_bytes(by):
+    """
+    Generates X random bytes, returns byte-string
+    :param by:
+    :return:
+    """
+    return random.bytes(by)
```
jpochyla: Are the wrappers necessary?

src/apps/monero/xmr/common.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Returns val if is not None, default instead
+    :param val:
+    :param default:
+    :return:
+    """
+    return val if not is_empty(val) else default
+
+
+def chunk(arr, size=1):
```
jpochyla: Isn't `trezor.utils.chunks` the same function?

src/apps/monero/protocol/key_image_sync.py
ph4r05: I was wondering, do you prefer Py3-only classes? Extending object is my habit from writing Py23 compatible code but I think in Py3 it is not needed, right?

src/apps/monero/diag.py
ph4r05: It is definitely not needed for the production. I used it quite often during the development process, e.g., to track memory consumption, memory leaks, heap fragmentation. So for testing / development it is nice to have. Depends what you prefer, whether to remove it from the code completely or wrap in `__debug__` branch.

src/apps/monero/controller/iface.py
```diff
@@ -0,0 +1,231 @@
...
+        await ui.backlight_slide(BACKLIGHT_DIM)
+        slide = ui.backlight_slide(BACKLIGHT_NORMAL)
+        # await ui.backlight_slide(BACKLIGHT_NORMAL)
+
+        text = Text("Signing transaction", ui.ICON_SEND, icon_color=ui.BLUE)
+        text.normal("Signing...")
+
+        try:
+            layout = await layout.simple_text(text, tm=500)
+            log.debug(__name__, "layout: %s", layout)
```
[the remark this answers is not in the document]
ph4r05: Makes sense, I will go through the code and fix it

src/apps/monero/controller/iface.py
```diff
@@ -0,0 +1,231 @@
...
+        await layout.require_confirm_tx(
+            self.gctx(ctx), addr.decode("ascii"), dst.amount, is_change
+        )
+
+    async def confirm_payment_id(self, payment_id, ctx=None):
+        """
+        Confirm payment ID
+        :param payment_id:
+        :param ctx:
+        :return:
```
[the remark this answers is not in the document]
ph4r05: Sure I will remove params docstring. It is mostly unused anyway (PyCharm is generating it automatically when adding the method comment). 
So the basic method description stays if is helpful enough (I mean not just rephrasing the method name)? 

Regarding type hints it is a bit more complicated if the param is of a complex type. I can add the hinting for trivially typed params but I don't think it is doable to add import-needed types as it pollutes general namespace with the imports. And to my surprise, the memory consumption of imported module is quite non-trivial (up to several kB).

src/apps/monero/controller/iface.py
```diff
@@ -0,0 +1,231 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
ph4r05: Yes, no problem.

src/apps/monero/controller/iface.py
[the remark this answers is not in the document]
prusnak: Micropython does the right thing and does not execute type hints (so they structures used there don't need to be imported). AFAIK we use the following trick:

```
if False:
   from some_module import complex_structure

def foo(bar: complex_structure):
    pass
```

That way it does not add any memory requirements to code (because type hints are not executed and `if False` is optimized out) and at the same time we can use mypy to do proper type-checking, because of some magic @jpochyla discovered. :D

src/apps/monero/protocol/key_image_sync.py
prusnak: Right. Also micropython is Py3 only, so it makes sense.

ph4r05:
I will address remaining comments soon

src/apps/monero/xmr/sub/addr.py
```diff
@@ -64,28 +64,16 @@ def classify_subaddresses(tx_dests, change_addr):
 def addr_eq(a, b):
     """
     Address comparisson. Allocation free.
+    :param a:
+    :param b:
+    :return:
```
tsusanka: Please don't add comments that provide exactly zero additional information

ph4r05:
Trezor-crypto related: `range_proof.{c,h}` and the corresponding unit test can be now removed as I deprecated C-implementation of the Borromean range proof due to heap fragmentation problems. Python implementation is almost as fast as the C implementation (using in-place operations).

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1415 @@
...
+        from trezor.messages.MoneroTransactionInitAck import MoneroTransactionInitAck
+        from trezor.messages.MoneroTransactionRsigData import MoneroTransactionRsigData
+
+        rsig_data = MoneroTransactionRsigData(offload_type=self.rsig_offload)
+        return MoneroTransactionInitAck(
+            in_memory=False,
+            many_inputs=True,
+            many_outputs=True,
+            hmacs=hmacs,
+            rsig_data=rsig_data,
```
tsusanka: Mmm, so we're returning `MoneroTransactionInitAck` with `MoneroTransactionRsigData` just to communicate `self.rsig_offload`? Also `in_memory=False, many_inputs=True, many_outputs=True` is hardcoded but might not be in the future?

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1508 @@
...
+
+            return Failure(code=FailureType.ActionCancelled, message="rejected")
+
+        gc.collect()
+        self._mem_trace(3)
+
+        # Basic transaction parameters
+        self.input_count = tsx_data.num_inputs
+        self.output_count = len(tsx_data.outputs)
+        self.output_change = misc.dst_entry_to_stdobj(tsx_data.change_dts)
```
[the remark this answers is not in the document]
ph4r05: We had a conversation with @jpochyla about removing StdObj and using protob directly once `FIELDS` are moved to a method `get_fields()`

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1508 @@
...
+        self.mixin = tsx_data.mixin
+        self.fee = tsx_data.fee
+        self.account_idx = tsx_data.account
```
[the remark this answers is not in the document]
ph4r05: It is needed to use subaddresses, `get_primary_change_address()` method uses it. Monero sends this information in unsigend_txes structure in the cold wallet tsx sign protocol.

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1415 @@
...
+        self.multi_sig = tsx_data.is_multisig
+        self.state.inp_cnt()
+        self.check_change(tsx_data.outputs)
```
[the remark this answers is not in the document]
ph4r05: Yep this is how Monero does that in cold wallet sign protocol. I presume there are scenarios where change transaction is not set - maybe sweep transactions (dummy change tsx). 

These structures are built by Monero when generating unsigned transaction set and I would rather not modify this logic. My code just translates it to the protobuf message.

src/apps/monero/xmr/sub/addr.py
```diff
@@ -0,0 +1,86 @@
...
+
+def public_addr_encode(pub_addr, is_sub=False, net=NetworkTypes.MAINNET):
+    """
+    Encodes public address to Monero address
+    """
+    net_ver = net_version(net, is_sub)
+    return encode_addr(net_ver, pub_addr.spend_public_key, pub_addr.view_public_key)
+
+
+def classify_subaddresses(tx_dests, change_addr):
```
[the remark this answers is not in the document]
ph4r05: Hmm I don't exactly follow. What do you mean? This is a direct port of `classify_addresses` method from Monero.

src/apps/monero/protocol/tsx_sign_builder.py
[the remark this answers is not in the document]
tsusanka: Oh right, I wrote this one before that. Ignore.

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1415 @@
...
+        self.need_additional_txkeys = num_subaddresses > 0 and (
+            num_stdaddresses > 0 or num_subaddresses > 1
+        )
+        self._mem_trace(4, True)
+
+        # Extra processing, payment id
+        self.tx.version = 2
+        self.tx.unlock_time = tsx_data.unlock_time
+        self.process_payment_id(tsx_data)
+        await self.compute_sec_keys(tsx_data)
```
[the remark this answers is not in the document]
ph4r05: Correct

src/apps/monero/protocol/tsx_sign_builder.py
ph4r05: I decided to group all rsig-related data to the `MoneroTransactionRsigData` as there are more possible offloading techniques. In this case, I found it acceptable as the message has only one field filled in so I found the overhead negligible.

`in_memory=False, many_inputs=True, many_outputs=True` are like 2 days old optimizations as we discussed to simplify the protocol and strip down the optimizations in order to unify the flow. 

In general, I am hesitant to modify messages structures too often because it usually requires PR and merging which is needed so I can continue with the work.

We can mark `in_memory=False, many_inputs=True, many_outputs=True` as deprecated and remove it in the next PR. What do you propose to do with `MoneroTransactionRsigData`?

src/apps/monero/xmr/sub/addr.py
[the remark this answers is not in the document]
tsusanka: I was just wondering why is the `addr_set` used and it seems that is used to skip duplicate addresses in tx_dests, isn't it?

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1415 @@
...
+        gc.collect()
+
+        # Iterative tx_prefix_hash hash computation
+        self._tprefix_update()
```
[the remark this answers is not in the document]
tsusanka: Indeed.

src/apps/monero/protocol/tsx_sign_builder.py
tsusanka: > We can mark in_memory=False, many_inputs=True, many_outputs=True as deprecated and remove it in the next PR. 

Sounds good. We can make an issue if you want to track those things.

> What do you propose to do with MoneroTransactionRsigData?

Let's keep it for now. It just seemed a bit weird to nest a MoneroTransactionRsigData in a MoneroTransactionInitAck, which does not contain any Rsig data obvisouly, because it's a first message. But it's a nit-pick, let's keep it for now.

ph4r05:
Btw just a minor thing, after using the latest trezor-common the src Travis job started failing. Ideas? I thought it may be good to fix it before merging protobuf PR with `get_fields`, maybe?

```
1.89s$ test "$GOAL" != "src" || pipenv run make templates_check
./tools/build_templates --check
--- ./tools/../src/apps/common/coininfo.py	2018-09-17 15:11:06.743812764 +0000
+++ /tmp/tmp.CucSLkIfYY	2018-09-17 15:12:11.042755779 +0000
@@ -520,7 +520,7 @@
         address_type_p2sh=5
         maxfee_kb=10000000
         signed_message_header="Namecoin Signed Message:\n"
-        xpub_magic=0x019da462
+        xpub_magic=0x0488b21e
         xpub_magic_segwit_p2sh=None
         xpub_magic_segwit_native=None
         bech32_prefix=None
make: *** [templates_check] Error 1
```

tsusanka:
@ph4r05 run `./tools/build_templates`

ph4r05:
After `pip install mako munch` I've got:

```
  File "./../vendor/trezor-common/tools/cointool.py", line 706, in <module>
    cli()
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "./../vendor/trezor-common/tools/cointool.py", line 702, in render
    do_render(file, dst)
  File "./../vendor/trezor-common/tools/cointool.py", line 675, in do_render
    render_file(src, dst, defs, support_info)
  File "./../vendor/trezor-common/tools/cointool.py", line 131, in render_file
    template = mako.template.Template(filename=src)
  File "/Users/dusanklinec/.pyenv/versions/3.6.5/lib/python3.6/site-packages/mako/template.py", line 283, in __init__
    "of the root path." % self.uri)
mako.exceptions.TemplateLookupException: Template uri "../src/apps/common/coininfo.py.mako" is invalid - it cannot be relative outside of the root path.
```

tsusanka:
Mmmm, maybe `cd tools; ./build_templates`?

ph4r05:
Aah I was originally in the `tools/` dir, now I tried `./tools/build_templates` and that worked.

src/apps/monero/xmr/sub/addr.py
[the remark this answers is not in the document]
ph4r05: Correct, original implementation also uses set to count unique addresses

src/apps/monero/protocol/tsx_sign_builder.py
```diff
@@ -0,0 +1,1416 @@
...
+            crypto.point_eq(
+                C,
+                crypto.point_add(
+                    crypto.scalarmult_base(mask), crypto.scalarmult_h(amount)
+                ),
+            ),
+            "OutC fail",
+        )
+
+    def _check_bproof(self, batch_size, rsig, masks):
```
tsusanka: I think this function is not being used?

ph4r05: Correct, it can be removed, the offloaded bulletproof is sent without commitment values so there is no check needed.

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+    state.multi_sig = tsx_data.is_multisig
+    check_change(state, tsx_data.outputs)
+    state.exp_tx_prefix_hash = tsx_data.exp_tx_prefix_hash
+
+    # Rsig data
+    state.rsig_type = tsx_data.rsig_data.rsig_type
+    state.rsig_grp = tsx_data.rsig_data.grouping
+    state.rsig_offload = state.rsig_type > 0 and state.output_count > 2
+    state.use_bulletproof = state.rsig_type > 0
+    state.use_simple_rct = state.input_count > 1 or state.rsig_type != 0
```
tsusanka: I'd like to make it here a little bit clearer. There are four options how to make the range signature:
1. borromean offloaded
1. borromean in trezor
1. bulletproof offloaded
1. bulletproof in trezor

Am I correct? And "simple" means in trezor, "full" offloaded?

If the above is correct `rsig_type` is equal to 0 if borromean and 1 if bulletproofs?

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+    # Provided tx key, used mostly in multisig.
+    if len(tsx_data.use_tx_keys) > 0:
+        for ckey in tsx_data.use_tx_keys:
+            crypto.check_sc(crypto.decodeint(ckey))
+
+        state.tx_priv = crypto.decodeint(tsx_data.use_tx_keys[0])
+        state.tx_pub = crypto.scalarmult_base(state.tx_priv)
+        state.additional_tx_private_keys = [
+            crypto.decodeint(x) for x in tsx_data.use_tx_keys[1:]
+        ]
```
tsusanka: If we're removing multisig, may I remove this?

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+            crypto.decodepoint(single_dest_subaddress.spend_public_key), state.tx_priv
+        )
+
+    state.need_additional_txkeys = num_subaddresses > 0 and (
+        num_stdaddresses > 0 or num_subaddresses > 1
+    )
+    state.mem_trace(4, True)
+
+    # Extra processing, payment id
+    state.tx.version = 2
```
tsusanka: Why 2? Current Monero transaction format?

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+
+    lextra = len(extra_nonce)
+    if lextra >= 255:
+        raise ValueError("Nonce could be 255 bytes max")
+
+    extra_buff = bytearray(3 + lextra)
+    extra_buff[0] = 2
+    extra_buff[1] = lextra + 1
+    extra_buff[2] = extra_prefix
+    utils.memcpy(extra_buff, 3, extra_nonce, 0, lextra)
```
tsusanka: Imho we can use simple `extra_buff[2:] = extra_nonce` here, because we are not slicing the `extra_nonce` (therefore no extra allocation). cc @jpochyla

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+    Payment id -> extra
+    """
+    if common.is_empty(tsx_data.payment_id):
+        return
+
+    from apps.monero.xmr.sub import tsx_helper
+    from trezor import utils
+
+    if len(tsx_data.payment_id) == 8:
```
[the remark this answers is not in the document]
ph4r05: Correct, 8B is encrypted payment ID, 32B is unencrypted payment ID.  
https://github.com/monero-project/monero/blob/83d8f03c2338e865ce72f4e2d46d43a48530ab10/src/wallet/wallet2.cpp#L4291

src/apps/monero/protocol/signing/step_01_init_transaction.py
jpochyla: 👍

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+
+    else:
+        raise ValueError("Payment ID size invalid")
+
+    lextra = len(extra_nonce)
+    if lextra >= 255:
+        raise ValueError("Nonce could be 255 bytes max")
+
+    extra_buff = bytearray(3 + lextra)
+    extra_buff[0] = 2
```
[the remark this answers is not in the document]
ph4r05: TX_EXTRA_NONCE == 2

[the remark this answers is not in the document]
ph4r05: TX_EXTRA_NONCE
`bool add_extra_nonce_to_tx_extra(std::vector<uint8_t>& tx_extra, const blobdata& extra_nonce)`

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+
+    writer = get_keccak_writer()
+    await protobuf.dump_message(writer, tsx_data)
+    writer.write(crypto.encodeint(state.tx_priv))
+
+    state.key_master = crypto.keccak_2hash(
+        writer.get_digest() + crypto.encodeint(crypto.random_scalar())
+    )
+    state.key_hmac = crypto.keccak_2hash(b"hmac" + state.key_master)
+    state.key_enc = crypto.keccak_2hash(b"enc" + state.key_master)
```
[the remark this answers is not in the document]
ph4r05: Correct, now it can be removed from the state

src/apps/monero/protocol/signing/step_01_init_transaction.py
ph4r05: Simple RCT is a standalone mode how MGsig ring is constructed from mixin. Used iff num_inputs == 1.

Borromean offloading is not currently implemented as we don't need it now. All other 3 options are used. 

rsig_type 0 = borromean, 1 = bulletproof

src/apps/monero/protocol/signing/step_01_init_transaction.py
ph4r05: Needed in step 6

src/apps/monero/protocol/signing/step_01_init_transaction.py
ph4r05: Yes, current transaction version in the Monero.

src/apps/monero/protocol/signing/step_01_init_transaction.py
```diff
@@ -0,0 +1,242 @@
...
+    if common.is_empty(tsx_data.payment_id):
+        return
+
+    from apps.monero.xmr.sub import tsx_helper
+    from trezor import utils
+
+    if len(tsx_data.payment_id) == 8:
+        view_key_pub_enc = tsx_helper.get_destination_view_key_pub(
+            tsx_data.outputs, state.change_address()
+        )
```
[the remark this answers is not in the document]
ph4r05: Can be done with exceptions but this is how Monero does it. Any chance `dst.addr.view_public_key` can be NULL_KEY_ENC? I have no clue.

src/apps/monero/protocol/signing/step_01_init_transaction.py
ph4r05: To be more precise, >=1 is Bulletproof.

```
enum RangeProofType { RangeProofBorromean, RangeProofBulletproof, RangeProofMultiOutputBulletproof, RangeProofPaddedBulletproof };
```

https://github.com/monero-project/monero/blob/83d8f03c2338e865ce72f4e2d46d43a48530ab10/src/ringct/rctTypes.h#L234

src/apps/monero/protocol/signing/step_01_init_transaction.py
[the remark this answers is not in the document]
tsusanka: Oooh, I finally get it, it has two levels: TX_EXTRA_NONCE -> TX_EXTRA_NONCE_PAYMENT_ID or TX_EXTRA_NONCE_ENCRYPTED_PAYMENT_ID.

So the tags described in my first comment are incorrect. All good :+1:

src/apps/monero/protocol/signing/step_01_init_transaction.py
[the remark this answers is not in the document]
tsusanka: Got it now, see above.

src/apps/monero/protocol/signing/step_01_init_transaction.py
tsusanka: Ok, let me summarize. There are two basic types of Ring Confidential Transactions (RCT):

1. RCTTypeFull
2. RCTTypeSimple

`state.use_simple_rct` denotes which one of those two is used.

For those two transaction types we have two (well, three) range proof types:
1. borromean
2. bulletproof in trezor
3. bulletproof offloaded

Is that correct?

tsusanka: Btw the [Zero to Monero](https://getmonero.org/library/Zero-to-Monero-1-0-0.pdf) book says it is the opposite, doesn't it?

> For multi-input transactions, existing Monero wallets use the RCTTypeSimple..

> Our perception is that the decision to limit RCTTypeFull transactions to one input..

ph4r05: Thats correct, source code has it right. Simple is for multi-inputs.

ph4r05: And I think the Full version is only used with Borromean rsig

src/apps/monero/protocol/signing/step_01_init_transaction.py
tsusanka: So in general the whole block after `# Provided tx key, used mostly in multisig.` deals with multisig? I'd remove it as discussed

ph4r05: Correct, it can be removed

src/apps/monero/protocol/signing/step_01_init_transaction.py
tsusanka: Ok, I think I'll just put it in a separate function for now and we can tackle it later to make it a little bit clearer.

ph4r05:
Hmm I think there is a small issue with generated mocks at the moment.
The `trezorcrypto` mock is one file but maybe it should actually be a package which contains submodules like `monero`.

Why: A function `add256_modm` defined in `modtrezorcrypto-monero.h` is in the generated mock in `trezorcrypto.add256_modm`, not in `trezorcrypto.monero.add256_modm` as it should be. 

A Quick & dirty hack is to class-scope all packages in `modtrezorcrypto`:

```python
class monero:

    @staticmethod
    add256_modm(a, b): 
        """Scalar addition"""

    class Ge25519:
        """EC point""""
```

Now the mock import / typing should work.

src/apps/monero/protocol/signing/step_06_set_out1.py
```diff
@@ -0,0 +1,361 @@
...
+
+    if (
+        state.current_input_index + 1 != state.input_count
+    ):  # todo check state.state.is_input_vins() - needed?
+        raise ValueError("Invalid number of inputs")
+
+    state.current_output_index += 1
+    state.mem_trace(2, True)
+
+    if dst_entr.amount <= 0 and state.tx.version <= 1:
```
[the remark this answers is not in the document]
ph4r05: ok sure

tsusanka:
@ph4r05 just pushed smth. So it's a go from my side

PR #296, opened 2018-08-04
display_identify: improve identification of ILI9341

mcudev:
I found another ID register in the datasheet for the ILI9341. 0xD3 is not mentioned in the ST7789V datasheet. That is why I only read the register if we don't already have a positive identification from the 0x04 register.

tested on prototype board + st7789v and with dev kit + ili9341.

### DIFF
--- a/embed/extmod/modtrezorui/display-stm32.h
+++ b/embed/extmod/modtrezorui/display-stm32.h
@@ -44,8 +44,7 @@
 
 // ST7789V  => 00 85 85 52
 // GC9307   => 00 00 93 07
-// ILI9341V => 00 00 80 00
-// some of the modules return "00 00 00 00", which means the vendor did not set the value
+// ILI9341V => 00 00 93 41
 static uint32_t __attribute__((unused)) display_identify(void)
 {
     volatile uint8_t c;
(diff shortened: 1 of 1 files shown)


prusnak:
Nice! Thanks!

prusnak:
Did small related refactor in bc5dba751a0a0078efd33fe1550222087823f0a6

mcudev:
i like that refactor! thanks!

prusnak:
And finally the runtime detection of connected display controller: 3022cd50c80975c1cc4a812e3d5e079cb9665eb7

Comments welcome!

mcudev:
will review. do you have a manual or datasheet for GC9307 anywhere? couldn't find one.

prusnak:
Sent via email

mcudev:
review:
- [x] no more extra build flags and recompiling, yay!
- [x] all references to DISPLAY_ILI9341V, DISPLAY_ST7789V, DISPLAY_GC9307 are removed
- [ ] need to re-identify for each code stage to set DISPLAY_ID since it gets re-initialized every time to 0 (.bss section variable).
- [ ] may want to update hardware.md to: `Driver IC: ST7789V, GC9307, or ILI9341V (on-chip display data RAM of 240x320x18 bits)`

PR #311, opened 2018-08-14
src/apps/ethereum: support full 32bit chain_id

hackmod:
* remove chain_id restriction to support full 32bit chain_id.
 * for chain_id > MAX_CHAIN_ID(2147483630) case, simply return v signature parity.
 * see also https://github.com/trezor/trezor-mcu/pull/399

### DIFF
--- a/src/apps/ethereum/sign_tx.py
+++ b/src/apps/ethereum/sign_tx.py
@@ -138,7 +138,9 @@ async def send_signature(ctx, msg: EthereumSignTx, digest):
 
     req = EthereumTxRequest()
     req.signature_v = signature[0]
-    if msg.chain_id:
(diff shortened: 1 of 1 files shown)


src/apps/ethereum/sign_tx.py
```diff
@@ -138,7 +138,9 @@ def get_total_length(msg: EthereumSignTx, data_total: int) -> int:
 
     req = EthereumTxRequest()
     req.signature_v = signature[0]
-    if msg.chain_id:
+    if msg.chain_id > MAX_CHAIN_ID:
```
prusnak: This doesn't work if `msg.chain_id` is None

prusnak:
Thanks!

PR #323, opened 2018-08-19
add Ontology support

backslash47:
Hi,

I prepared signing support for Ontology based on the messages in trezor-common. Currently there are 6 methods defined:
getPublicKey, getAddress, signTransfer, signWithdrawOng, signOntIdRegister and signOntIdAddAttributes

Please review the code, if it is alright or something needs to be changed.

I have also prepared tests in python-trezor but I don't know how to build debug build of trezor-core and run in emulator, so I can use LoadDevice in the test.


prusnak:
To run emulator in debug mode, just run `PYOPT=0 ./emu.sh`

backslash47:
Thanks, will finish the tests and make PR to python-trezor

backslash47:
Hello, I added the python-trezor PR here https://github.com/trezor/python-trezor/pull/301

backslash47:
Hi,
Your assumption is correct. There is no concept of change. About the bip32 path: in ledger integration we use 44'/1024'/0'/0/x 
It will be good to have the same paths. 
And because lot of people holding ont is also holding neo and because there is currently a swap from nep-5 to main net tokens, the path could also be 44'/888'/0'/0/x where 888 is NEO coin. I called it NEO compatible addresses in my wallet (Cyano wallet - https://github.com/OntologyCommunityDevelopers/cyano-walle)

tsusanka:
> Your assumption is correct. There is no concept of change. About the bip32 path: in ledger integration we use 44'/1024'/0'/0/x

Ok, great.

> And because lot of people holding ont is also holding neo and because there is currently a swap from nep-5 to main net tokens, the path could also be 44'/888'/0'/0/x where 888 is NEO coin. I called it NEO compatible addresses in my wallet (Cyano wallet -OntologyCommunityDevelopers/cyano-wallet )

Uff, sorry I don't know much about Ontology and NEO. So you can store NEO coins on a an Ontology address?

backslash47:
Let me explain:
The address format is the same. ONT holders are in many cases also NEO holders.
Ontology came to existence first as a NEP-5 token (similar to ERC20 token on Ethereum).
This year a main net was launched and a swap procedure from NEP-5 token to main net token is undergoing.

And yes, basically you can store NEO and ONT on the same addresses, it is just on different network.

backslash47:
Hello,

Anything else needs to be improved?

prusnak:
We are in process of stabilizing the new firmware release, which will be released in the following weeks. Once this is done, we'll review your pull request.

backslash47:
Thanks for the information

matejcik:
@prusnak the point of `setup.cfg` line is to _enable_ the tests for this PR - so that we see results of tests that are xfailed by default

so please keep that line.

backslash47:
Hi,

about the writers:
There is missing write_uint16, write_bool (can be replaced by write_uint8), write_varint (specific to NEO and Ontology) and write_bytes_with_length (also specific to NEO and Ontology, because it uses write_varint). But I can move the missing methods to common writers if you want.

backslash47:
@prusnak I simplified the message registration, but I can not test it, because the build is failing due to some cardano inconsistencies.

prusnak:
Rebase your work on top of the master and it should work (don't touch trezor-crypto and trezor-common submodules).

backslash47:
I have it already rebased on master. Please have a look at https://travis-ci.org/trezor/trezor-core/jobs/428480375 starting from line 1969, that is exactly the same error I encountered locally.

prusnak:
In https://github.com/trezor/trezor-core/pull/323/commits/e576f7a1230d0c4f977637bf339ddfbcdf873153 you change both trezor-common and trezor-crypto submodules. I told you not to do it. If you fix this (remove this change from the commit) it will work.

backslash47:
My mistake. Will try it without it.

backslash47:
All should be ready.

backslash47:
@prusnak Hello, I am wondering, the next step to add Ontology integration (after firmware) would be trezor-connect? Or is there something what should be done before.

prusnak:
Yes, Connect.

backslash47:
Hi, 
what is the timetable for merging this PR please? Is there anything I can do to help with it?

src/apps/ontology/sc/native_builder.py
```diff
@@ -0,0 +1,118 @@
...
+    builder.write_push_bytes(ret, bytes(funcName, "ascii"))
+    builder.write_push_hexstr(ret, contract)
+    builder.write_push_int(ret, 0)
+    writer.write_byte(ret, opcode.SYSCALL)
+    builder.write_push_bytes(ret, b"Ontology.Native.Invoke")
+
+    return ret
+
+
+def _write_native_code_script(ret: bytearray, arr: list) -> None:
```
matejcik: this seems more complicated than it needs to be

the usage in your code is for exactly two cases:
1. `params` is a list of a single ParamStruct, and
2. `params` is a list of a list of a single ParamStruct

also, in all cases the ParamStruct seems to not be recursive, i.e., it is always a list of simple values

I propose we convert this code so that it _always_ takes a list of simple values, and an additional argument which controls whether the input should be considered an array of n elements
so let's say:
```python
def _write_native_code_script(ret: bytearray, params: list, array_len: int = None):
    write_byte(NEWSTRUCT)
    (...)
    for val in params:
        if isinstance(...):
            (...)
        (...)
    
    write_byte(FROMALTSTACK)
    if array_len is not None:
        write_push_int(array_len)
```

This won't be a generic script builder anymore, but we don't seem to need that, and it makes the code simpler to reason about. It will also let us get rid of `ParamStruct` and pass the lists directly.

Or is there a reason to keep the code generic, that I'm not seeing?

src/apps/ontology/sc/opcode.py
```diff
@@ -0,0 +1,151 @@
+# OpCode constants
```
matejcik: we don't need a full list of opcodes. Please reduce this only to opcodes that are actually used in the code.

src/apps/ontology/sign_ont_id_add_attributes.py
```diff
@@ -0,0 +1,37 @@
...
+)
+from trezor.messages.OntologySignOntIdAddAttributes import (  # isort:skip
+    OntologySignOntIdAddAttributes
+)
+
+
+async def sign_ont_id_add_attributes(ctx, msg: OntologySignOntIdAddAttributes):
+    await _require_confirm(ctx, msg.transaction, msg.ont_id_add_attributes)
+
+    address_n = msg.address_n or ()
```
matejcik: please don't allow empty `address_n` and use path validation

src/apps/ontology/sign_ont_id_register.py
```diff
@@ -0,0 +1,33 @@
...
+
+from .layout import require_confirm_ont_id_register
+from .serialize import serialize_ont_id_register
+from .sign import sign
+
+
+async def sign_ont_id_register(ctx, msg: OntologySignOntIdRegister):
+    await _require_confirm(ctx, msg.transaction, msg.ont_id_register)
+
+    address_n = msg.address_n or ()
```
matejcik: please don't allow empty `address_n` and use path validation

src/apps/ontology/sign_transfer.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .layout import require_confirm_transfer_ong, require_confirm_transfer_ont
+from .serialize import serialize_transfer
+from .sign import sign
+
+
+async def sign_transfer(ctx, msg: OntologySignTransfer):
+    await _require_confirm(ctx, msg.transaction, msg.transfer)
+
+    address_n = msg.address_n or ()
```
matejcik: please don't allow empty `address_n` and use path validation

src/apps/ontology/sign_withdraw_ong.py
```diff
@@ -0,0 +1,28 @@
...
+
+from .layout import require_confirm_withdraw_ong
+from .serialize import serialize_withdraw_ong
+from .sign import sign
+
+
+async def sign_withdraw_ong(ctx, msg: OntologySignWithdrawOng):
+    await _require_confirm(ctx, msg.transaction, msg.withdraw_ong)
+
+    address_n = msg.address_n or ()
```
matejcik: please don't allow empty `address_n` and use path validation

matejcik (changes requested):
I did a review of your submission. please respond to comments and modify code where appropriate.

Most notably: Please implement path validation and never allow empty `address_n`.

Also, as noted before, please get rid of (most of) `writer.py`, instead reuse `apps.common.writers` and add the missing calls

src/apps/ontology/writer.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Writes boolean
+    """
+    if n:
+        write_byte(w, 1)
+    else:
+        write_byte(w, 0)
+
+
+def write_varint(w: bytearray, n: int):
```
matejcik: you might want to use `write_varint` from `apps.wallet.sign_tx.writers`

src/apps/ontology/get_address.py
```diff
@@ -0,0 +1,25 @@
...
+from trezor.messages.OntologyAddress import OntologyAddress
+
+from .helpers import get_address_from_public_key
+
+from apps.common import seed
+from apps.common.layout import show_address, show_qr
+
+
+async def get_address(ctx, msg):
+    address_n = msg.address_n or ()
```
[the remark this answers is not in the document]
matejcik: ah of course because when this PR was first submitted, it was like this everywhere.

well, please see:
https://github.com/trezor/trezor-core/blob/master/src/apps/ethereum/get_address.py#L12
https://github.com/trezor/trezor-core/blob/master/src/apps/ethereum/address.py#L38

matejcik:
In https://github.com/trezor/trezor-core/pull/323#issuecomment-415752419 you said that the bip32 path is m/44'/1024'/0'/0/**i**
That seems rather weird, given that the last part of the path is non-hardened. You're not exporting the public nodes so it doesn't seem to affect you in practice, but it's still weird.

It would be better if we could prefer paths like m/44'/1024'/**a**', and allow the /0/0 suffix for compatibility. However, we don't know the wallet landscape. It seems OK on Cyano wallet, where it seems there are no path-based accounts? But, are there wallets that iterate the last part for accounts instead of the usual account part?

prusnak:
The work is continued here: https://github.com/trezor/trezor-firmware/pull/177

PR #329, opened 2018-08-23
Unify common writer logic

jpochyla:
Serialization logic from altcoin and bitcoin apps was moved to `apps.common.writers`. Also:

- `write_*` functions always take the writer as first param
- `serialize_*` functions always return bytes

### DIFF
--- a/src/apps/common/writers.py
+++ b/src/apps/common/writers.py
@@ -1,0 +1,68 @@
+def empty_bytearray(preallocate: int) -> bytearray:
+    """
+    Returns bytearray that won't allocate for at least `preallocate` bytes.
+    Useful in case you want to avoid allocating too often.
+    """
+    b = bytearray(preallocate)
+    b[:] = bytes()
+    return b
+
+
+def write_uint8(w: bytearray, n: int) -> int:
+    assert 0 <= n <= 0xFF
+    w.append(n)
+    return 1
+
+
+def write_uint32_le(w: bytearray, n: int) -> int:
+    assert 0 <= n <= 0xFFFFFFFF
+    w.append(n & 0xFF)
+    w.append((n >> 8) & 0xFF)
+    w.append((n >> 16) & 0xFF)
+    w.append((n >> 24) & 0xFF)
+    return 4
+
+
+def write_uint32_be(w: bytearray, n: int) -> int:
+    assert 0 <= n <= 0xFFFFFFFF
+    w.append((n >> 24) & 0xFF)
+    w.append((n >> 16) & 0xFF)
+    w.append((n >> 8) & 0xFF)
+    w.append(n & 0xFF)
+    return 4
+
+
+def write_uint64_le(w: bytearray, n: int) -> int:
+    assert 0 <= n <= 0xFFFFFFFFFFFFFFFF
+    w.append(n & 0xFF)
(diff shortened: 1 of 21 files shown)


tsusanka:
(tests are failing on style)
```
src/apps/stellar/writers.py:1:1: F401 'ustruct' imported but unused
src/apps/wallet/sign_tx/writers.py:5:1: F401 'apps.common.writers.empty_bytearray' imported but unused
```

src/apps/common/writers.py
```diff
@@ -0,0 +1,64 @@
+def empty_bytearray(preallocate: int) -> bytearray:
+    b = bytearray(preallocate)
```
tsusanka: Great rename. Could you maybe add a short note on why we're doing this?

src/apps/common/writers.py
```diff
@@ -0,0 +1,64 @@
...
+    b[:] = bytes()
+    return b
+
+
+def write_uint8(w: bytearray, n: int) -> int:
+    assert n >= 0 and n <= 0xFF
```
tsusanka: I'm not sure it's worth it, but we can simplify these expressions to `assert 0 <= n <= 0xFF`

src/apps/nem/mosaic/serialize.py
```diff
@@ -98,8 +94,10 @@ def _write_property(w: bytearray, name: str, value):
             value = "false"
     elif type(value) == int:
         value = str(value)
-    elif type(value) != str:
+    if type(value) != str:
```
tsusanka: I think this is a bug. The `elif` should not be dropped

src/apps/nem/transfer/serialize.py
```diff
@@ -20,52 +25,54 @@ def serialize_transfer(
     payload: bytes = None,
     encrypted: bool = False,
 ) -> bytearray:
-    tx = write_common(
+    tx = serialize_tx_common(
         common,
-        bytearray(public_key),
+        public_key,
```
tsusanka: serialize_tx_common expects [bytearray](https://github.com/trezor/trezor-core/pull/329/files#diff-86afd8d197d0e672da6c3180a7bf85e1R8) but here the `public_key` argument is defined as bytes. Is this native conversion okay?

jpochyla: If we use bytearray only for reading, bytes are OK.

src/apps/nem/mosaic/serialize.py
jpochyla: The rest of the function depends on `value` being `str`, it was a bit clearer for me this way. But I don't really have any strong preference.

src/apps/common/writers.py
jpochyla: 👍

src/apps/common/writers.py
jpochyla: Yep!

src/apps/nem/mosaic/serialize.py
tsusanka: Oh, now I got it. It's fine, I haven't realised all the "conversions" are to `str`.

src/apps/nem/transfer/serialize.py
tsusanka: :+1:

PR #331, opened 2018-08-24
Unify wire registering of wire callbacks and load app modules automatically

jpochyla:


prusnak:
Maybe rename `wire.add` to `wire.register` and `wire.register` to `wire._register`? Or do we use original `wire.register` without the `add` wrapper somewhere?

prusnak:
utACK

PR #336, opened 2018-08-27
Mako templates for coins, tokens, etc.

matejcik:
with updated `trezor-common`, handy `make templates` that render every found `foo.bar.mako` to the corresponding `foo.bar`, and `make templates_check` for Travis to make sure that you're not commiting changes in the autogenerated files


tsusanka:
utACK

I might test tomorrow. The concept is great though

prusnak:
Thanks!

PR #337, opened 2018-08-28
List of used derivation paths

tsusanka:
This documents tries to summarize what paths should be provided for which coin. If another path is provided, trezor shows a warning. Few notes:

- If NEM's _correct_ path is `44'/43'/a'`, we should allow it as well, shouldn't we? Not just the _historical_ one, similar as we do for Ethereum. And maybe the same goes for Ripple? 
- I'm putting this into docs/ folder and I think we might move all the coin's README here (as @prusnak suggested already before).

@prusnak @jpochyla @matejcik @szymonlesisz could you have a look if this seems fine?

### DIFF
--- a/docs/coins/README.md
+++ b/docs/coins/README.md
@@ -1,0 +1,37 @@
+
+List of used BIP-44 derivation paths
+------------------------------------
+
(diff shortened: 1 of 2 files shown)


matejcik:
lgtm

prusnak:
Merged! Thx

PR #340, opened 2018-08-30
Remove split_words

### LINKED ISSUE #330: Investigate utils.split_words
Investigate utils.split_words and its potentional merge with the Text component. It is currently used by NEM only and maybe it doesn't have to be?

jpochyla:
`split_words` and the `Text` component implement the same layout algorithm.

Fixes #330

### DIFF
--- a/src/apps/common/signverify.py
+++ b/src/apps/common/signverify.py
@@ -1,9 +1,7 @@
 from ubinascii import hexlify
 
-from trezor import ui
 from trezor.crypto.hashlib import sha256
-from trezor.ui.text import TEXT_MARGIN_LEFT
-from trezor.utils import HashWriter, chunks, split_words
+from trezor.utils import HashWriter
 
 from apps.wallet.sign_tx.signing import write_varint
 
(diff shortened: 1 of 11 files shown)


tsusanka:
So does this mean we can remove split_words? Because I don't see the remove

jpochyla:
Updated, sorry.

tsusanka:
This looks good, but I still don't like the API - probably same as you :). I tried to come up with a better one, but yes it's tricky. How about:

```python
class WordsText(Text):
    def __init__(...):
        super().__init__(..., new_lines=False)

    def words(self, content: str):
        self.normal(*content.split(" "))
```

and then simply `t = WordsText(...)` and `t.words('some words')`. But I've realized this sucks as well, it is really unintuitive which one to use (Text or WordsText). I guess the correct solution is to have the `new_lines=True` as default, but I understand it is breaking change.

So the conclusion is we can merge this one and revisit again as part of the UI redesign if needed.

tsusanka:
@jpochyla let's merge this? It's not perfect, but I see it as an improvement

jpochyla:
Yes same here, let's merge it and revisit in the future.

PR #354, opened 2018-09-15
Add Tron Support

fbsobreira:


prusnak:
Don't define your own writers in `apps/tron/serialize.py` but rather reuse ones from `apps/common/writers.py`

fbsobreira:
Ok, thanks... I will work on it!

fbsobreira:
@prusnak  I have removed the writers that I was not using and also revert to prior trezor-common commit. Please let me know if there is any other suggestions.
Thank you

fbsobreira:
@prusnak I removed Tron Messages from the PR, as they are already merged into trezor-core. Please let me know if there is anything else I should do.

Thank you.

matejcik:
let's reuse our built-in protobuf serialization capability instead of typing it out explicitly.

we probably don't want to use the protobuf definitions from trezor's protocol, because a) they (deliberately) don't match the Tron protobufs, and b) they should be able to evolve independently from Tron.

instead, please describe the parts of protobufs that you use in the same format as our protobuf generator uses, e.g., https://github.com/trezor/trezor-core/blob/master/src/trezor/messages/TezosDelegationOp.py, and place them into a submodule of your `tron` app.
You can probably avoid writing the `__init__` method, just `get_fields` with appropriate data should be sufficient.

then you can create a `BufferWriter` class that lets you call `protobuf.dump_message` and retrieve the result, instead of explicitly writing out the serialization field by field

src/apps/tron/get_address.py
```diff
@@ -0,0 +1,24 @@
...
+from trezor.messages.TronAddress import TronAddress
+
+from .helpers import get_address_from_public_key
+
+from apps.common import seed
+from apps.common.layout import show_address, show_qr
+
+
+async def get_address(ctx, msg):
+    address_n = msg.address_n or ()
```
matejcik: in all cases please use address validation, see e.g. Ethereum:
https://github.com/trezor/trezor-core/blob/master/src/apps/ethereum/get_address.py#L12
https://github.com/trezor/trezor-core/blob/master/src/apps/ethereum/address.py#L38

matejcik:
closing for cleanup now.
if you resume work on this, please apply your changes to the monorepo at https://github.com/trezor/trezor-firmware, along with your python-trezor PR, and resubmit

PR #357, opened 2018-09-18
Protobuf message fields moved to get_fields method

ph4r05:
Related to:
- https://github.com/trezor/trezor-common/issues/171
- https://github.com/trezor/trezor-common/pull/212
- 2 commits, version bump + re-generated messages and templates, protobuf.py update


src/protobuf.py
```diff
@@ -204,8 +207,9 @@ def __init__(self):
         setattr(msg, fname, fvalue)
 
     # fill missing fields
-    for tag in msg.FIELDS:
-        field = msg.FIELDS[tag]
+    fields = msg.get_fields()
```
jpochyla: Redundant?

jpochyla: Nonono, I mean that `fields` already contains the message fields, since line 155.

ph4r05: 😅 true

jpochyla:
Thanks!

PR #358, opened 2018-09-20
decred: add initial support

matheusd:
Sending for early review to know whether the approach is acceptable for inclusion.

I've roughly followed what was done in trezor-mcu: adding a few conditional checks for decred then customizing where needed.

Please advise on whether this approach will be acceptable so I can do some clean up and improvements or if you'd prefer a more modular implementation (eg: creating a `sign_tx_decred`, `get_prev_output_value_decred`, etc).

### DIFF
--- a/src/apps/wallet/sign_tx/signing.py
+++ b/src/apps/wallet/sign_tx/signing.py
@@ -2,7 +2,7 @@ from micropython import const
 
 from trezor.crypto import base58, bip32, cashaddr, der
 from trezor.crypto.curve import secp256k1
-from trezor.crypto.hashlib import sha256
+from trezor.crypto.hashlib import blake256, sha256
 from trezor.messages import OutputScriptType
 from trezor.messages.TxRequestDetailsType import TxRequestDetailsType
 from trezor.messages.TxRequestSerializedType import TxRequestSerializedType
@@ -34,6 +34,11 @@ _BIP32_CHANGE_CHAIN = const(1)
 # the maximum allowed change address.  this should be large enough for normal
 # use and still allow to quickly brute-force the correct bip32 path
 _BIP32_MAX_LAST_ELEMENT = const(1000000)
+
+
+DECRED_SERIALIZE_FULL = const(0 << 16)
+DECRED_SERIALIZE_NO_WITNESS = const(1 << 16)
+DECRED_SERIALIZE_WITNESS_SIGNING = const(3 << 16)
 
 
 class SigningError(ValueError):
@@ -143,6 +148,14 @@ async def check_tx_fee(tx: SignTx, root: bip32.HDNode):
         elif not await confirm_output(txo, coin):
             raise SigningError(FailureType.ActionCancelled, "Output cancelled")
 
+        if coin.decred:
+            if txo.decred_script_version is not None and txo.decred_script_version != 0:
+                raise SigningError(
+                    FailureType.ActionCancelled,
+                    "Cannot send to output with script version != 0",
+                )
+            txo_bin.decred_script_version = txo.decred_script_version
+
         write_tx_output(h_first, txo_bin)
         hash143.add_output(txo_bin)
         total_out += txo_bin.amount
@@ -182,6 +195,19 @@ async def sign_tx(tx: SignTx, root: bip32.HDNode):
     tx_req = TxRequest()
     tx_req.details = TxRequestDetailsType()
     tx_req.serialized = None
+
+    h_prefix_sign = None
+    if coin.decred:
+        h_prefix_sign = HashWriter(blake256)
+
+        # used to validate no changes between check_tx_fee and phase 2
+        h_second = HashWriter(sha256)
+
+        # used to validate no changes between phase 2 and decred witness
+        h_third = HashWriter(sha256)
+        h_fourth = HashWriter(sha256)
+        write_uint32(h_prefix_sign, tx.version | DECRED_SERIALIZE_NO_WITNESS)
+        write_varint(h_prefix_sign, tx.inputs_count)
 
     for i_sign in range(tx.inputs_count):
         progress.advance()
@@ -234,7 +260,11 @@ async def sign_tx(tx: SignTx, root: bip32.HDNode):
             key_sign = node_derive(root, txi_sign.address_n)
             key_sign_pub = key_sign.public_key()
             hash143_hash = hash143.preimage_hash(
-                coin, tx, txi_sign, ecdsa_hash_pubkey(key_sign_pub), get_hash_type(coin)
+                coin,
+                tx,
+                txi_sign,
+                ecdsa_hash_pubkey(key_sign_pub, coin),
+                get_hash_type(coin),
             )
 
             # if multisig, check if singing with a key that is included in multisig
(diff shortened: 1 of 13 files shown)


src/apps/wallet/sign_tx/signing.py
```diff
@@ -183,6 +185,14 @@ class SigningError(ValueError):
     tx_req.details = TxRequestDetailsType()
     tx_req.serialized = None
 
+    decred = coin.curve_name == 'secp256k1-decred' # TODO: maybe improve this detection method
```
prusnak: I pushed the change to master where you can just use `coin.decred`, so drop this line completely.

src/apps/wallet/sign_tx/signing.py
```diff
@@ -183,6 +185,14 @@ class SigningError(ValueError):
...
+    h_prefix_sign = None
+    if decred:
```
prusnak: And use `coin.decred` here instead of `decred`.

src/apps/wallet/sign_tx/signing.py
```diff
@@ -259,6 +269,22 @@ class SigningError(ValueError):
 
             tx_req.serialized = tx_ser
 
+        elif decred:
```
prusnak: And use `coin.decred` here instead of `decred`.

prusnak:
I put some inline comments, please address them, finish `# TODO check if input changed` and I think we are good to go!

prusnak:
Also if you know how to extract long chunks of code in the signing loop into separate functions, such as `get_prev_output_value_decred` please do so.

matheusd:
Addressed the review items, plus added a restriction to spending to/from decred_script_version != 0 as was done in trezor-mcu.

Long term, I think it might be better to add decred as a separate app (similar to tezos, etc) so that we'll support staking, changes to signature hash scheme, other signature types, etc. But this should suffice for initial support.

prusnak:
What wallet are you using to test the integration with?

matheusd:
dcrwallet+decrediton and dcrwallet+[my custom decred/trezor cli wallet](https://github.com/matheusd/trezor-decred-poc).

It will be slightly tricky if you want to try and get your hands into this to test because it needs custom dcrwallet and decrediton builds, but I can either arrange binaries or help you out if you want to try and build yourself.

prusnak:
Merged via 7834d06aac60bd37af9981a4f5b914ba1555b455

Thank you!

prusnak:
@matheusd I just noticed that your code behaves differently then the code on Trezor One - so this test fails: https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_signtx_decred.py 

Please advise.

matheusd:
I'm trying to setup my dev env to run the tests in python-trezor. 

But looking at the test, maybe the problem was the change of the master key we introduced in `trezor-crypto`?

matheusd:
Ok, so I've ran the test and reviewed things. Given that the protocol implemented for T2 is slightly different, the easiest thing would be to implement a new test specifically for T2.

Would that be acceptable?

prusnak:
> Would that be acceptable?

I would prefer to have the same protocol for T1 and T2. Why is your protocol for T2 different? Why did you implement it in a different fashion?

matheusd:
> Why did you implement it in a different fashion than on T1?

To minimize the amount of changes to the existing signing.py.

For example, `check_tx_fee()` in T2 returns a hash (h_first) but that's not the same as the prefix hash needed by decred, so I have to request the initial tx data again inside `sign_tx()`.

`singing.c` in T1 uses global variables for all state, so it was easier to just add an additional hash (`hash_prefix`) and modify the appropriate places to fill it, which I assumed wouldn't be acceptable for T2.

If having the same protocol for decred is a hard requirement, I add further changes to support it, no problem.

prusnak:
@matheusd I understand the motivation now. However, when I remove all calls to `self.client.set_expected_responses` in `trezorlib/tests/device_tests/test_msg_signtx_decred.py` from python-trezor, the test will still fail with error in the emulator:

```
  File "/home/stick/work/trezor/trezor-core/src/trezor/wire/__init__.py", line 127, in session_handler
  File "/home/stick/work/trezor/trezor-core/src/trezor/wire/__init__.py", line 124, in session_handler
  File "/home/stick/work/trezor/trezor-core/src/trezor/wire/__init__.py", line 162, in protobuf_workflow
  File "/home/stick/work/trezor/trezor-core/src/trezor/wire/__init__.py", line 149, in protobuf_workflow
  File "/home/stick/work/trezor/trezor-core/src/trezor/ui/__init__.py", line 125, in inner
  File "/home/stick/work/trezor/trezor-core/src/trezor/ui/__init__.py", line 122, in inner
  File "/home/stick/work/trezor/trezor-core/src/apps/wallet/sign_tx/__init__.py", line 38, in sign_tx
  File "/home/stick/work/trezor/trezor-core/src/apps/wallet/sign_tx/__init__.py", line 28, in sign_tx
  File "/home/stick/work/trezor/trezor-core/src/apps/wallet/sign_tx/signing.py", line 482, in sign_tx
TypeError: unsupported types for __gt__: 'NoneType', 'int'
```

The problematic line contains:
```
            if txi.amount > authorized_in:
```

but the `txi` object does not contain `amount` field. This is weird - the amount field for inputs should be enabled only for segwit-enabled coins and we have segwit: False for decred.

I am reverting the decred code in master for now. And I created a `decred` branch where we can work on the integration further. I'll disable the decred test (`test_msg_signtx_decred.py`) in python-trezor, but please re-enable it in your local checkout (remove skip_t2 directive) and try it with `self.client.set_expected_responses` removed from the code.

matheusd:
Removed the offending line, fixed multisig (which I wasn't testing because our wallet doesn't support bip48) and sign/verify.

I'll also submit a PR to python-trezor adding a test for sign/verify that should pass in both T1 and T2.

A note about the offending line in the previous comment: decred does support specifying the input amount irrespective of segwit, though currently it's not included in the signed hash. It does need to be either pushed by the client or filled as 0 by `write_tx_input_decred_witness` since it needs to be transmitted. Our wallet client code was filling it correctly, that's why the python test was failing in this particular line.

prusnak:
It seems the changes are going into the right direction, but you managed to break other signing tests. Please revisit your changes and try to fix all unit test breakages.

prusnak:
I picked up your changes into the `decred` branch and squashed them into a single commit.

Now when I run `test_msg_signtx_decred.py` (after removal of skip_t2 attribute) I see the following error in the emulator. The good thing is that other signtx tests are not broken this time.

```
1539096516439618 trezor.wire WARNING failure: Encountered invalid prev_hash
```

prusnak:
I was just looking more into the unit test and realized that T2 signing, as suggested by you, has more messages in the signing exchange. I think it would be better if we follows T1 way more closely as it is more efficient when it comes to number of messages. "Global" state is not a problem if it just about introducing more hashers. You already introduced 3 of them anyway.

prusnak:
Merged via 2277a9c7541c4ed7113471d93d62c4c93f546ca2

Thank you Matheus! :tada:

matheusd:
Thank you for all the help and patience in getting this in!

PR #359, opened 2018-09-23
Stellar - Fix typo in confirm_init

zulucrypto:


prusnak:
thanks! although singing is also good :D

PR #371, opened 2018-10-10
unix: user profiles support for emulator

malinah:
Support for user profiles allowing switching of emulator configurations.


prusnak:
Merged via fc9bc85ea3444a59e1737d9d5d93a83c6171344f

PR #373, opened 2018-10-10
firmware: log exceptions in debug mode (no silent unhandled exceptions)

malinah:
If the exceptions are not handled as part of the program logic at least log them. Now you can find out when you run out of memory etc.

### DIFF
--- a/src/boot.py
+++ b/src/boot.py
@@ -1,4 +1,4 @@
-from trezor import config, loop, res, ui
+from trezor import config, log, loop, res, ui
 from trezor.pin import pin_to_int, show_pin_timeout
 
 from apps.common.request_pin import request_pin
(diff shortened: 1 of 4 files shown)


src/boot.py
```diff
@@ -18,7 +18,8 @@
                     return
                 else:
                     label = "Wrong PIN, enter again"
-        except:  # noqa: E722
+        except Exception as e:  # noqa: E722
```
prusnak: remove `# noqa: E722`, not needed anymore

src/boot.py
```diff
@@ -18,7 +18,8 @@
...
+            log.exception(__name__, e)
```
prusnak: Maybe put this in `if __debug__:`

src/boot.py
```diff
@@ -18,7 +18,8 @@
...
             pass
```
prusnak: Remove `pass`, not needed anymore

jpochyla:
❤️

PR #395, opened 2018-11-02
xmr: key image sync progress info

ph4r05:
Adds very simple progress view for key image sync so user knows Trezor is working on the sync (otherwise it could look like something went wrong / frozen)


src/apps/monero/layout/confirms.py
```diff
@@ -123,3 +123,13 @@
     text = Text("Signing transaction", ui.ICON_SEND, icon_color=ui.BLUE)
     text.normal(*info)
     text.render()
+
+
+@ui.layout
+async def keyimage_sync_step(ctx, current, total_num):
+    if current is None:
+        return
+    info = ["Signing inputs", "%d/%d" % (current + 1, total_num)]
```
tsusanka: mmm, "signing inputs"?

ph4r05: 😅somehow I missed it

PR #396, opened 2018-11-06
WIP: Refactor Bitcoin-like transaction signing

jpochyla:


jpochyla:
Deferred, current commits moved into #411.

PR #407, opened 2018-11-09
added CoinInfo for Actinium (ACM)

brakmic:
Expanded *COINS* array in coininfo.py.

I've already checked the protos in *trezor-common* but couldn't find anything that should be adapted to Actinium. However, there is a high probability that I have missed some important parts, so please, correct me if I am wrong.

I am currently reading the docs on API, Testing, and Emulation in *trezor-core*. 

Regards,

### DIFF
--- a/src/apps/common/coininfo.py
+++ b/src/apps/common/coininfo.py
@@ -65,6 +65,26 @@ class CoinInfo:
 # fmt: off
 COINS = [
     CoinInfo(
+        coin_name="Actinium",
+        coin_shortcut="ACM",
+        address_type=53,
+        address_type_p2sh=55,
+        maxfee_kb=40000000,
+        signed_message_header="Actinium Signed Message:\n",
+        xpub_magic=0x0488b21e,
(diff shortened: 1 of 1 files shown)


matejcik:
the file you edited is autogenerated.

you will need to add the appropriate data to trezor-common/defs, update reference in trezor-core and regenerate this file with `make templates`

matejcik:
seeing that ACM is already merged in trezor-common, all left to do is the steps outlined above

brakmic:
Many thanks for the explanation! 👍

matejcik:
Submodule references in Git point to a specific commit by design. What your `git clone --recursive` does is fetch _that particular commit_. It's not supposed to get the latest version; on the contrary, _you're_ supposed to update the reference so that it points to a commit with your defs in it. (usually to the most up-to-date version, through e.g. going to the `vendor/trezor-common` subdir and `git checkout master; git pull`)

brakmic:
Many thanks. Already changing it!

PR #417, opened 2018-11-20
Cardano improvements

refi93:
- unified address export screen with other cryptourrencies (displaying QR/string)
- changed UI of signing transactions to be more similar to BTC
- better grouping of information when signing Tx and dropping info about network and tx size which we deem redundant


refi93:
We thought a bit more about the "network" attribute being sent with the transaction and we suggest changing it directly to protocol magic - the integer being serialized in the signature, specific for the network. This allows trezor to sign transactions for custom networks. There are already PRs on trezor-common and trezor-connect repos to update the name of the field in the calls: https://github.com/trezor/trezor-common/pull/244 , https://github.com/trezor/connect/pull/272

To prevent an attack where the adversary tricks the signer into taking transaction inputs from the mainnet when sending a testnet transaction we suggest displaying the network explicitly. It's probably a hypothetical attack, but better stay on the safe side.

src/apps/cardano/sign_tx.py
```diff
@@ -118,39 +110,29 @@
...
-        if network == 1:
-            self.network_name = "Testnet"
-            self.network_magic = b"\x01\x1a\x41\x70\xcb\x17\x58\x20"
-        elif network == 2:
-            self.network_name = "Mainnet"
-            self.network_magic = b"\x01\x1a\x2d\x96\x4a\x09\x58\x20"
-        else:
-            raise wire.ProcessError("Unknown network index %d" % network)
+
+        self.network_name = KNOWN_PROTOCOL_MAGICS.get(protocol_magic) or 'Unknown'
```
ppershing: Unless micrypython differs from python here, we should be able specify default value in `get` like this:

```
>>> {'xx': 1}.get('xx', 'def')
1
>>> {'xx': 1}.get('yy', 'def')
'def'
```

prusnak: This works in uPy.

src/apps/cardano/sign_tx.py
```diff
@@ -7,69 +7,61 @@
...
+    return True
+
 
 async def show_tx(
     ctx,
     outputs: list,
     outcoins: list,
-    change_derivation_paths: list,
-    change_coins: list,
     fee: float,
```
ppershing: Hmm, why is fee `float` here?

refi93: good point, it's an int actually, fixed

src/apps/cardano/sign_tx.py
refi93: fixed

src/apps/cardano/layout/__init__.py
```diff
@@ -3,11 +3,67 @@
...
+        to_pages = list(chunks(to_lines[1:], LINES_PER_PAGE))
+        for page in to_pages:
+            t = Text("Confirm transaction", ui.ICON_SEND, icon_color=ui.GREEN)
+            for line in page:
+                t.bold(line)
+            pages.append(t)
+
+    paginator = paginate(create_renderer(ConfirmDialog), len(pages), const(0), pages)
+    return await ctx.wait(paginator) == CONFIRMED
+    return True
```
jpochyla: Dead code.

refi93:
fixed code style and removed dead code. I did not manage to run the tests in python-trezor though: https://github.com/trezor/python-trezor/pull/342 seems like emulation of swipe and confirm does not work for test_msg_cardano_sign_transaction.py. It waits for manually swiping and confirming. And I did not manage to understand from the code, how to properly emulate those events. What am I missing?

jpochyla:
@refi93 We are testing the swipable dialogs this way: https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_resetdevice_t2.py#L56, it should work in all cases that use `trezor.ui.scroll`.

prusnak:
FYI I just merged https://github.com/trezor/trezor-common/pull/244 into master

refi93:
I added ButtonRequests to `confirm_transaction` and `confirm_sendind` screens so it is testable, hence the force push.

refi93:
updated trezor-common submodule

refi93:
I recently added output address validation, to solve an issue we recently came across - i.e. that a specially constructed invalid address may cause the raw transaction to have more outputs than advertised by trezor ui.

refi93:
I recently added output address validation

PR #438, opened 2018-12-16
Added Gincoin support

bedeabza:


prusnak:
This we'll do automatically before the release of next firmware.

bedeabza:
Ok, was not familiar with the process. What is the ETA and how will I know of the updates?

Issue #402, opened 2018-11-08
Trezor Model T flicker screen after I upgraded v2.0.9


0x5826:
Today I upgraded my trezor T to v2.0.9.
when I touch the screen, Trezor Model T flicker screen and i can't click exactly. no matter i downgrade the version or not, it doesn't work. but in bootloader mode, there is no problem.

I don't think The screen of model T have problems, because it works fine before I upgrade. and some people got the same problem like me.

seem like  #104 
ps:the video is https://www.youtube.com/watch?v=V6oyu-kfD9M

matejcik:
FWIW I was able to reproduce the video perfectly just moments ago, but now I can't anymore :(

So just to note: there is also something to do with the touch panel having problems around screen edges. A workaround for the issue seen on video is moving your thumb a little up.

Flickering when touching elsewhere than the button is also present in lower versions, maybe a little less but it's definitely there.

0x5826:
I upgraded v2.0.10, I tested it but the issue always happen on macbookpro macOS 10.14.2.
check the video [https://youtu.be/cbt5-aXAhCw](url)
Is it my laptop usb voltage problem or other problem?

prusnak:
It's really weird this happens only on your MacBook, but desktop PC is fine. We'll be tuning sensitivity of the touchscreen in the upcoming release, so this might get fixed too.

prusnak:
@0x5826 we'll handle this in https://github.com/trezor/trezor-core/pull/439

PR #439, opened 2018-12-16
Find ideal value for touch screen sensitivity (aka threshold)

prusnak:
It is possible to set touch screen sensitivity. The default seems to be 0x12 (I read this value from 0x80 register), but maybe a lower value is a better choice. 

See the example of touch screen sensitivity issue here:  https://github.com/trezor/trezor-core/issues/402 (and video inside: https://youtu.be/cbt5-aXAhCw)

**Do not merge until we find out the ideal value!**

### DIFF
--- a/embed/prodtest/main.c
+++ b/embed/prodtest/main.c
@@ -224,6 +224,33 @@ static void test_touch(const char *args)
     touch_power_off();
 }
 
+static void test_sensitivity(const char *args)
+{
+    int v = atoi(args);
+
+    touch_power_on();
+    touch_sensitivity(v & 0xFF);
+
+    display_clear();
+    display_refresh();
+
+    for (;;) {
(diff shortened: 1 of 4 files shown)


prusnak:
I have just checked the new sensitivity value on regular Model T and it seems OK.

(I use reset_device workflow as a test, because it is possible to use it immediately after uploading the firmware and it contians regular confirm button, hold to confirm button and also swiping - all 3 usecases are good for testing sensitivity).

@jpochyla  Please merge if you feel the results are the same for you too.

prusnak:
I've just added modified prodtest to the commit. New command "SENS" can test sensitivity setting.

prusnak:
merged in https://github.com/trezor/trezor-core/commit/4dc8110b3103e395c10a7f4df6278bef49eb1463

PR #445, opened 2018-12-29
Add Unobtanium (UNO) Support 

Flurbos:
More info is available in the trezor-common PR _(link below)_
**Unobtanium trezor-common PR request:**  trezor/trezor-common#257


prusnak:
We'll update the submodule once the trezor-common PR will go in.

PR #452, opened 2019-01-09
Add sha256sum check to arm-none-abi toolchain binaries in Dockerfile

TheCharlatan:
The docker build fails if the hash is incorrect.


prusnak:
This won't work if a developer changes TOOLCHAIN_FLAVOR to `src`.

prusnak:
Also please update to a new release `gcc-arm-none-eabi-8-2018-q4-major` while you are at it.

TheCharlatan:
Force-pushed. Now uses gcc and downloads to a generic filename.

prusnak:
That was not what I had in mind. I will rework the patch by myself. Thanks for the idea.

PR #455, opened 2019-01-11
add Komodo support

ca333:
This PR adds support for Komodo (https://komodoplatform.com)

Komodo Transactions need the nLockTime to be set with a recent timestamp in order for the outputs to be eligible for reward claiming hence we did `timestamp - 777` and modified the class SigningError to differentiate between zcash and komodo overwintered/sapling TX.


ca333:
We are adding the definition to trezor-common. Will update this PR with reformated komodo.py.

ca333:
Thanks for the hint @prusnak  - will do

ca333:
please let me know if all is good now @prusnak

prusnak:
Rebase on top of current master, remove merge commits (`git rebase master`)

ca333:
Komodo has a reward feature which allows a `utxo>=10KMD` to collect the reward [1] (summing up to ~5.1 % p.a. - blockchain enforced "interest/reward"). Our implementation does check against the set nLockTime which has to be explicitly set in the past since the "interest/reward" checks would otherwise prevent the utxo from accumulating the bonus[2]. So we could technically also do an offset of 500 or 1000. However, we traditionally have been using "777" and also refered to "777" in other similar cases[3]. Hardcoding this value would be a big help automatically preventing any potential issue.

**[1]** https://komodoplatform.atlassian.net/wiki/spaces/KPSD/pages/71729215/What+is+the+5+Komodo+Stake+Reward
**[2]** https://github.com/komodoplatform/komodo/blob/master/src/komodo_interest.h#L101
**[3]** https://github.com/LedgerHQ/ledger-nano-s/issues/21

prusnak:
However, my question was, whether we can keep the Trezor logic unchanged and just change the value that is passed to Trezor from the caller (software wallet) - in another word - the caller would decrease the value.

ca333:
@prusnak Yes, we can pass the decreased locktime via our wallet.

prusnak:
In that case, none of these changes are needed and this commit: https://github.com/trezor/trezor-common/commit/a37e80905e33d619ea82921c671be189d9c7971e will enable KMD in the next firmware release.

I suggest you check BRANCH_ID here:

https://github.com/trezor/trezor-core/blob/master/src/apps/wallet/sign_tx/zcash.py#L81

and here:

https://github.com/trezor/trezor-core/blob/master/src/apps/wallet/sign_tx/zcash.py#L126

Whether they are the same in Komodo and open a pull request if not.

Also I recommend you to add unit test similar to this one to python-trezor:

https://github.com/trezor/python-trezor/blob/master/trezorlib/tests/device_tests/test_msg_signtx_zcash.py

ca333:
Thanks a lot @prusnak - yes i have confirmed that our branch IDs are identical with zcash, so all is good. I will add python device tests this week. Looking forward to the next firmware release.

PR #467, opened 2019-01-23
Andrewkozlik/trezor storage integration

andrewkozlik:
Integrate the new trezor-storage module into trezor-core.

PR #469, opened 2019-01-24
Trezor storage integration rebased

prusnak:
supersedes https://github.com/trezor/trezor-core/pull/467


tests/test_trezor.config.py
```diff
@@ -41,7 +41,7 @@ def test_wipe(self):
         self.assertEqual(v1, None)
 
     def test_lock(self):
-        for _ in range(128):
+        for i in range(128):
```
prusnak: i is not used, unnecessary change

andrewkozlik: Agreed. I added that while debugging and forgot to clean up.

prusnak:
Merged via 6d817a813d584a9a5240ad261e0cd64febd9eabf

Thanks!

PR #473, opened 2019-02-12
tezos: add staking support

adonagy:
Added staking support - signing endorsements and baked block headers - for tezos.

The signing process is automatic after the user confirms entering staking mode on the device. While in this mode, the device is rejecting all messages except the ones used for staking and others like initialize, wipe device, etc.. To leave staking mode the device ask for PIN. 

Depends on https://github.com/trezor/trezor-common/pull/267


src/apps/tezos/layout.py
```diff
@@ -62,10 +62,39 @@
...
     return chunks(address, 18)
 
 
 def format_tezos_amount(value):
     formatted_value = format_amount(value, TEZOS_AMOUNT_DIVISIBILITY)
     return formatted_value + " XTZ"
+
+
+def get_operation(wm):
+    if wm is 1:
```
Adman: More pythonic way is to write it this way 
```python
return "Block" if wm == 1 else "Endorsement"
```

Also, `is` is used for object identities (https://stackoverflow.com/questions/13650293/understanding-pythons-is-operator)

src/apps/tezos/writers.py
```diff
@@ -0,0 +1,24 @@
...
+    write_uint64_be,
+)
+
+write_uint8 = write_uint8
+write_uint32 = write_uint32_be
+write_uint64 = write_uint64_be
+write_bytes = write_bytes
+
+
+def write_bool(w: bytearray, boolean: bool):
```
Adman: This function is already in `sign_tx.py`. It would be cool to move all writers from `sign_tx.py` to `writers.py`, so we do not have the same function defined multiple times :)

matejcik: yes, please do that :)

src/apps/tezos/writers.py
```diff
@@ -0,0 +1,24 @@
...
+    if boolean:
+        write_uint8(w, 255)
+    else:
+        write_uint8(w, 0)
+
+
+# write uint16 in be
+def write_uint16(w: bytearray, n: int):
```
matejcik: also please move this to `apps.common.writers`

adonagy:
Thank you @matejcik and @Adman for the comments. I will work on changes ASAP

src/apps/tezos/control_baking.py
```diff
@@ -0,0 +1,29 @@
...
+from trezor.pin import pin_to_int
+
+from apps.common.request_pin import request_pin
+from apps.tezos import helpers, layout
+
+
+async def control_baking(ctx, msg):
+
+    if not config.has_pin():
+        await layout.no_pin_dialog(ctx)
```
matejcik: this should probably return Failure (or raise an error) straight away, see comments below

src/apps/tezos/sign_baker_op.py
```diff
@@ -0,0 +1,83 @@
...
+    node = keychain.derive(msg.address_n, helpers.TEZOS_CURVE)
+
+    if helpers.check_baking_confirmed():
+        # Accept endorsements and block headers only, otherwise lock the device
+        if _check_operation_watermark(msg.magic_byte):
+            w = bytearray()
+            _get_operation_bytes(w, msg)
+            sig_prefixed = await _sign(ctx, bytes(w), node, msg)
+        else:
+            await helpers.prompt_pin()
```
matejcik: I'm not sure about this part. It will just throw you out of the staking mode when you send an "invalid" operation? Why not have the same behavior when you send a non-staking instruction?

My gut instinct would be to raise a wire error but stay in staking mode. But I'd like to talk a little more of the security implications.
ISTM that the only reason for sending something weird *in a baking operation* is a wire error, in which case the proper thing to do is retry.
OTOH someone might be trying to bust the one allowed operation?
But as long as they're busting, they could bust the other-operation-protection.
I dunno.

matejcik:
I'm still thinking about whether we need a way to disable the staking mode at run-time (as opposed to restarting the device). The ability to exit staking mode complicates doing this securely.

My current idea is this: when you receive "enable staking" message, you unregister *all* protobuf handlers, and enter an infinite loop inside the tezos app. This loop would only await the Baker op and process it forever. Any other message would cause a failure, which could either be handled within the app, or if you managed to shoot the app down, Trezor would not be able to process any more messages.

given that this veers deep into implementation details, i'm also ccing @jpochyla :)

src/apps/homescreen/homescreen.py
```diff
@@ -45,6 +46,12 @@ def display_homescreen():
             ui.WIDTH // 2, 22, "PIN NOT SET!", ui.BOLD, ui.BLACK, ui.YELLOW
         )
         ui.display.bar(0, 30, ui.WIDTH, ui.HEIGHT - 30, ui.BG)
+    elif helpers.check_baking_confirmed():
```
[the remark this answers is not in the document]
adonagy: > Does it even make sense to do staking mode on a seedless device? Would it be important to show both? (probably yes)

It indeed does not make sense to do baking mode on a seedless device. And as you said, baking mode requires PIN. In our app we also guide the user to set up the device properly (backup, set PIN, etc.). So i think this message having low priority is OK.

src/apps/tezos/layout.py
```diff
@@ -62,10 +62,50 @@
...
+    text.normal(str(msg.block_header.level))
+    text.bold("Type:")
+    text.normal("Baking")
+    text.render()
+    ui.display.backlight(ui.BACKLIGHT_NORMAL)
+
+async def no_pin_dialog(ctx):
+    text = Text("Pin is not set", ui.ICON_WRONG, icon_color=ui.RED)
+    text.normal("You need to set a pin to start staking")
+    return await require_confirm(ctx, text, ButtonRequestType.Other)
```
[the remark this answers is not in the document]
adonagy: > this is not good UX: with require_confirm you tell the user to look on device, which tells them that they can't do a thing and asks yes/no.

Yes, that makes sense. I removed the dialog

src/apps/tezos/sign_baker_op.py
```diff
@@ -0,0 +1,83 @@
...
+)
+
+
+async def sign_baker_op(ctx, msg, keychain):
+    paths.validate_path(ctx, helpers.validate_full_path, path=msg.address_n)
+    node = keychain.derive(msg.address_n, helpers.TEZOS_CURVE)
+
+    if helpers.check_baking_confirmed():
+        # Accept endorsements and block headers only, otherwise lock the device
+        if _check_operation_watermark(msg.magic_byte):
```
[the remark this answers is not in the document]
adonagy: My primary goal was to implement identical message sent by the tezos node. Then I looked at the implementation of the tezos transactions by @Adman and he was indeed force setting all the values associated with the operation type. 
 https://github.com/trezor/trezor-core/blob/396f5f1937dd4de7bfcb20954a059175e31fa98e/src/apps/tezos/sign_tx.py#L102

So to keep the consistency I also force-set the `magic_byte` and `tag`.

src/apps/tezos/sign_baker_op.py
adonagy: > OTOH someone might be trying to bust the one allowed operation?
But as long as they're busting, they could bust the other-operation-protection.

After discussion, we concluded that this check is indeed redundant for the reason you mentioned above.

adonagy:
Thank you for the review. I've updated the code according the requested changes. 

> When you receive "enable staking" message, you unregister all protobuf handlers, and enter an infinite loop inside the tezos app.

We discussed it with @jurajselep and we think the proposed flow is indeed a better approach.

We now unregister all the unused (in tezos baking) protobuf handlers when the device receives the "enable staking" messages.

> So if the device in staking mode gets disconnected, you already need to enter the pin (and therefore perform user interaction)

To exit this mode you need to disconnect the device. As you mentioned above, the user is required to enter the pin after he plugs back the device. This also simplifies the workflow in our app by removing the need for a "stop_staking" operation.

matejcik:
re last commit: in Trezor only ed25519 is supported which does not have non-hardened path derivation
therefore only 3-component fully hardened path was allowed.

Is there any reason to allow a 4- or 5-component paths?

src/apps/tezos/helpers.py
```diff
@@ -1,8 +1,15 @@
 from micropython import const
 
+from trezor import config
```
matejcik: none of the changes in this file are needed anymore, istm?

src/apps/tezos/sign_tx.py
```diff
@@ -5,8 +5,13 @@
 from trezor.messages.TezosSignedTx import TezosSignedTx
 
 from apps.common import paths
-from apps.common.writers import write_bytes, write_uint8
+# from apps.common.writers import write_bytes, write_uint8
```
matejcik: it seems better to import these from `apps.common.writers` and only import `write_bool` from tezos?

src/apps/tezos/writers.py
matejcik: looking back at this, does it make sense to have a `writers.py` just for a `write_bool` function? maybe moving it e.g. to `helpers.py` would be better?

src/trezor/wire/__init__.py
```diff
@@ -209,3 +213,14 @@ def import_workflow(ctx, req, pkgname, modname, *args):
...
+
+def tezos_remove_handelrs():
+    # remove all handlers, except the ones used in tezos baking
+    for handler in workflow_handlers.keys():
+        if handler not in tezos_baking_allowed_messages:
+            workflow_handlers.pop(handler, None)
+
+
+def is_baking():
+    return len(workflow_handlers) == len(tezos_baking_allowed_messages)
```
matejcik: let's not do the check this way.

instead, i dunno, have a top-level `TEZOS_BAKING` constant that you set to `True` when entering baking mode.
but `homescreen.py` is probably more appropriate for it, given that it would be the only place that uses it.

(Also consider `workflow.closedefault` and `startdefault`. Perhaps making a custom homescreen layout would be a better solution? But I'm still unsure as to how exactly it would work. Just an option to keep in mind.)

src/apps/tezos/__init__.py
```diff
@@ -9,3 +9,5 @@ def boot():
     wire.add(MessageType.TezosGetAddress, __name__, "get_address", ns)
     wire.add(MessageType.TezosSignTx, __name__, "sign_tx", ns)
     wire.add(MessageType.TezosGetPublicKey, __name__, "get_public_key", ns)
+    wire.add(MessageType.TezosSignBakerOp, __name__, "sign_baker_op", ns)
```
matejcik: don't register this handler here at boot, instead register it when entering baking mode.

that way you don't need a check in `sign_baker_op`

adonagy:
> Is there any reason to allow a 4- or 5-component paths?
I've changed the validation so 3 and 4 component paths are allowed (all hardened). The reason being  is that we want to allow migration from ledger, which uses 4 component hardened paths.

src/apps/tezos/control_baking.py
```diff
@@ -0,0 +1,20 @@
...
+
+from apps.common.request_pin import request_pin
+from apps.tezos import helpers, layout
+
+
+async def control_baking(ctx, msg):
+    if not config.has_pin():
+        return Failure()
+
+    if msg.baking is True:
```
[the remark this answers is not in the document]
adonagy: > we can probably drop the baking field and leave the message empty, right?

Right, we can send an empty message now, that we omitted the stop_staking operation

src/apps/tezos/sign_baker_op.py
```diff
@@ -0,0 +1,87 @@
...
+    sig_prefixed = helpers.base58_encode_check(
+        signature, prefix=helpers.TEZOS_SIGNATURE_PREFIX
+    )
+
+    if msg.endorsement is not None:
+        await layout.show_endorsement_operation(msg)
+    elif msg.block_header is not None:
+        await layout.show_baking_operation(msg)
+
+    await ctx.wait(loop.sleep(4 * 1000 * 1000))
```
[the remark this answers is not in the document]
adonagy: > why this number?

This just lets the `show_baking_operation` and `show_endorsement_operation` screens to stay for 4 seconds. I could not find another way of showing an "informational" screen without the buttons. If we dont "sleep" the loop here, the screens just pops up for a small instant then switching back to the homescreen. 

Any suggestions how to handle this situation in some other way?

adonagy:
I also rebased the branch to be up to date with master.

src/trezor/wire/__init__.py
adonagy: I think making a custom homescreen layout would solve the issue with the `show_baking_operation` and `show_endorsement_operation` screens (need to sleep the loop to show for the defined time). We could use the homescreen to show the last signed baking operation. But I'm still not sure how to make this work.

adonagy:
The ability to stop staking is still in question. A baker, who has been delegated coins, needs to move those tokens frequently (to achieve this he needs to reconnect the device in the current flow). And having the ability to move tokens (transaction operations) in baking mode is not secure IMO. 

If we stay with the current flow, an alternative approach would be to get an another Trezor T device with the same seed and use that for transactions while the other one is handling the baking. 

Any thoughts?

adonagy:
last commit: I've added the option to sign the baking operation without displaying it on the device

matejcik:
closing for cleanup now.
please apply your changes to the monorepo at https://github.com/trezor/trezor-firmware, together with the corresponding trezor-common PR, and resubmit

PR #480, opened 2019-02-21
NumPad to enter a number from arbitrary range easily

tsusanka:
To easily enter numbers from an arbitrary range. I've decided to go with the design where as much buttons are present as possible, because then 32 (the max number of shares in SLIP-39) fits in three pages.

peek 2019-02-21 09-45

For a perfect UI maybe a back button should be present when you click some + button? So you can go back

### DIFF
--- a/src/trezor/ui/num_pad.py
+++ b/src/trezor/ui/num_pad.py
@@ -1,0 +1,47 @@
+from trezor import ui
+from trezor.ui import display
+from trezor.ui.button import BTN_CLICKED, Button
+
+ITEMS_PER_PAGE = 11
+
+
+def digit_area(i):
+    return ui.grid(i + 3)  # skip the first line
+
+
+class NumPad(ui.Widget):
+    def __init__(self, label: str, start: int, end: int):
+        """
+        Generates a numpad with numbers from `start` to `end` excluding.
+        """
+        self.label = label
+        self.start = start
+        self.end = end
+        self.page = 0
+
+        self._generate_buttons()
+
+    def render(self):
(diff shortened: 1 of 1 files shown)


src/trezor/ui/num_pad.py
```diff
@@ -0,0 +1,48 @@
...
+def digit_area(i):
+    return ui.grid(i + 3)  # skip the first line
+
+
+class NumPad(ui.Widget):
+    def __init__(self, label: str, start: int, end: int):
+        """
+        Generates a numpad with numbers from `start` to `end` excluding.
+        """
+        display.clear()
```
jpochyla: This shouldn't really be needed, `ui.layout` usually takes care of clearing the display.

src/trezor/ui/num_pad.py
```diff
@@ -0,0 +1,48 @@
...
+        for btn in self.buttons:
+            if btn.touch(event, pos) == BTN_CLICKED:
+                if "+" in btn.content:
+                    self.page += 1
+                    self._generate_buttons()
+                else:
+                    return btn.content
+
+    def _generate_buttons(self):
+        display.clear()
```
jpochyla: Please add a code comment explaining why we are clearing the screen here.

tsusanka: Hm, because we need to redraw the buttons with other values. I can comment, but it makes sense, doesn't it?

tsusanka: This is what I get if I don't:

emu00000001

The last row should not be there at all.

PR #481, opened 2019-02-21
docker: emulator support refacoring

ph4r05:
I was playing with the Trezor emulator support in docker a bit. I had some issues with the original Dockerfile (I needed also GUI) so I created a slightly modified version. 

Dockerized Trezor-core emulator is beneficial also to other integrating projects, e.g., I could make use of it in Monero CI: https://github.com/monero-project/monero/pull/4977

In the simplest version, the emulator runs without GUI, minimal setup for integration testing of the integration. The idea is to clone a trezor-core branch, build the docker image and run it. 

We could also upload master firmware releases to the Docker repository (testing against released versions).

It is also possible to build and run with GUI. It was tested on OSX, should work on Linux too, potentially also on Windows with an appropriate X server.

All options are in [docker/emulator/README.md](https://github.com/trezor/trezor-core/blob/646d75c7edd8ef6b63886eb79825ff48f9495c01/docker/emulator/README.md)

Give it a thought, if this is OK for you. 

Thanks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # initialize from the image
 
-FROM debian:9
+ARG BASE_IMAGE=debian:9
+FROM ${BASE_IMAGE} AS base
 
 ARG TOOLCHAIN_FLAVOR=linux
 ENV TOOLCHAIN_FLAVOR=$TOOLCHAIN_FLAVOR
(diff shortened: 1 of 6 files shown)


prusnak:
Moved to https://github.com/trezor/trezor-firmware/issues/91

PR #489, opened 2019-03-06
Add DEXON (DXN) support

boczeratul:
[DEXON](https://dexon.org) is a new blockchain with efficient consensus algorithm to provide
- high TPS (around 10k at the moment)
- low latency (about 1s at the moment)

It uses the same account system as Ethereum and is EVM compatible.

ChainID: `237`
Registered coin type index: `237` [ref: slip-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md#registered-coin-types)

### DIFF
--- a/src/apps/ethereum/networks.py
+++ b/src/apps/ethereum/networks.py
@@ -143,6 +143,13 @@ NETWORKS = [
         rskip60=False,
     ),
     NetworkInfo(
+        chain_id=237,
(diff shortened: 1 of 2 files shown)


prusnak:
New Ethereum network definition should be added here: https://github.com/trezor/trezor-common/tree/master/defs/ethereum

PR #491, opened 2019-03-07
u2f: capitalize Bitwarden

yegortimoshenko:
E.g. see https://twitter.com/bitwarden.


prusnak:
Their logo uses `bitwarden`, so let's not use capitalization.

PR #496, opened 2019-03-11
Store mnemonic as bytes

tsusanka:
This is one of the preliminary works needed to introduce SLIP-39. It introduces another layer between seed.py and storage.py called mnemonic.py (though maybe we could find a better name?). The mnemonic module operates on storage and decides if mnemonic is BIP-39 or SLIP-39 and returns values appropriately.

For BIP-39 the mnemonic is stored in bytes, which is basically the same as it used to be (more of a syntax thing). We could also store the "internal entropy" of it but since the seed is then derived from hashing the string form, this makes more sense.

This PR contains temporary commit which points to python-trezor/trezor-common tsusanka/pre-slip39 branches. If approved I will commit those to master and regenerate this commit.


src/apps/common/storage.py
```diff
@@ -44,6 +45,17 @@ def _get_bool(app: int, key: int, public: bool = False) -> bool:
     return config.get(app, key, public) == _TRUE_BYTE
 
 
+def _set_uint8(app: int, key: int, val: int):
+    config.set(app, key, val.to_bytes(1, "big"))
```
tsusanka: I used "big" because that's what I've found in other instances, but maybe we should put this in to a constant?

jpochyla:
Looks good, let's review `mnemonic.py` after we add SLIP-39.

Issue #282, opened 2018-07-21
Confidential Assets tracking issue


apoelstra:
It would be great to have support for Confidental Assets (CA), which is used in the Elements blockchain platform and in Liquid, the commercial sidechain based on Elements. I'm happy to help move this forward, and am available by email at apoelstra at blockstream.com for any clarification or guidance.

CA is described in this blog post: https://blockstream.com/2017/04/03/blockstream-releases-elements-confidential-assets.html
It is essentially the same as Confidential Transactions, which involves the use of Pedersen commitments in place of amounts, alongside rangeproofs, which I believe is already supported as part of Trezor's Monero support. CA mades one of the two EC generators variable and adds a new "asset surjection proof" which is described here: https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/surjection/surjection.md

We've implemented CA in full as part of libsecp256k1-zkp, our fork of the Bitcoin Core libsecp256k1 library.

For referenc, the API for making surjection proofs is here: https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/include/secp256k1_surjectionproof.h
and the core crypto functionality is here:
https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/surjection/main_impl.h
https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/surjection/surjection_impl.h

Cheers

apoelstra:
Hi @prusnak ,

Our current use of `malloc` is well-encapsulated in only one place: `secp256k1_context_create` and `secp256k1_context_free`. These functions need to be called only during system init/deinit and is designed to be easily replaceable on embedded systems that do not support malloc.

We also use malloc in our unit tests and benchmarks, which I assume doesn't matter.

So I think option (2) is actually likely to be the simpler one.

prusnak:
Is secp256k1-zkp a superset of libsecp256k1? If yes, I think it's worth
the effort integrating secp256k1 into our stack.

apoelstra:
Yes, secp256k1-zkp is a superset of libsecp256k1 - all the new code is in the `src/modules/` tree (and `Makefile.am` and `configure.ac` are correspondingly modified).

[...]

Issue #422, opened 2018-11-23
TT: 2.0.9, inconsistent ethereumVerifyMessage failure


szymonlesisz:
Trying to verify ethereum message, intentionally with invalid data.
I get: { error:"Firmware error", code:"Failure_FirmwareError" }

Looks like it's not consistent with T1 which in this case returns:
{ error:"Invalid signature", code:"Failure_DataError" }

PR #500, opened 2019-03-14
Add extmod for secp256k1-zkp

real-or-random:
This includes the https://github.com/ElementsProject/secp256k1-zkp library (which is a fork from  sipa/secp256k1 used in Bitcoin Core)  and uses it for signing transactions and signing and verifying messages for ECDSA on the secp256k1 curve.

The Ethereum app is left untouched. I assume because they use Keccak-256 as a message digest and not SHA256. @apoelstra 

This is the first step towards solving #282.

Some background to understand the open issues below:
secp256k1-zkp (and in fact sipa/secp256k1) relies on two precomputation tables to speed up signing and for verification.  secp256k1(-zkp) supports that the precomputed table values for the signing table are stored in the executable (by defining `USE_ECMULT_STATIC_PRECOMPUTATION`) and thus do not need to be created at running time.

It's WIP because the following is still open (and maybe more):
 -  [x] secp256k1-zkp uses two "error callbacks". These are called when some bad error occurs and we don't know how to continue except for shutting down the execution properly, see https://github.com/bitcoin-core/secp256k1/blob/77668c09da756190e188521d87aaaa08736fa196/include/secp256k1.h#L218 and the function below for details. The default callbacks use `abort()` and `fprintf()`, so we need to replace them. I replaced them by functions which throw Python exceptions. Is this the right thing to do on trezor?
 - [x] The size of the *verification* table is configurable. A larger table makes verification possibly faster. We set it to 4096 bytes (of RAM). Is this is fine for you as memory footprint? I guess even smaller is okay because verification speed is not so crucial. We need to see if we want to move that to CCMRAM too as mentioned in https://github.com/trezor/trezor-core/issues/282#issuecomment-467508352 .
 - [x] The table(s) are stored in a `secp256k1_context` which is currently created when it's first used. Is that okay or do we want to create at program startup?
 - [ ] This currently relies on a dev branch of secp256k1-zkp. We'll merge this into master once we know that we're on the right track.
- [x] We wanted to enable `USE_ECMULT_STATIC_PRECOMPUTATION` but it increases the executable by 64 KiB, which is about 17 KiB too much after we rebased on master. To make the build work I've added a HACK commit that disables `USE_ECMULT_STATIC_PRECOMPUTATION in the firmware build again. But that means calling into secp256k1-zkp functions will fail at runtime because we don't give it enough memory to create the table. Note that it's probably really better to build this in the executable because otherwise we need 64 KiB of RAM instead of flash. Also, we need to keep in mind that currently no rangeproof functions are used, so those will be removed from the executable. If we really call these, the size will increase further.
- [x] There is naming collision between with the header `secp256k1.h`. For some reason the build currently works (at least on my machine) but this is probably fragile. Our goal is to make all of this work without the need to vendor specific secp256k1-zkp changes, and we'd rather not want to rename our header. I think it's easiest to rename your currently used `secp256k1.h`? Is that fine for you? Otherwise we can also use #includes with directories.

We opened this mostly to get some feedback on the build system. We can also decide to use this PR just for the build system and handle the other stuff in a different PR if that works better for you. 

cc @romanz @apoelstra

### DIFF
--- a/SConscript.firmware
+++ b/SConscript.firmware
@@ -6,6 +6,7 @@ CCFLAGS_MOD = ''
 CPPPATH_MOD = []
 CPPDEFINES_MOD = []
 SOURCE_MOD = []
+SOURCE_MOD_SECP256K1_ZKP = []
 
 PYOPT = '1'
 
@@ -83,6 +84,32 @@ SOURCE_MOD += [
     'vendor/trezor-crypto/secp256k1.c',
     'vendor/trezor-crypto/sha2.c',
     'vendor/trezor-crypto/sha3.c',
+]
+
+# libsecp256k1-zkp
+CPPPATH_MOD += [
+    'vendor/secp256k1-zkp',
+    'vendor/secp256k1-zkp/src',
+    'vendor/secp256k1-zkp/include',
+]
+CPPDEFINES_MOD += [
+    'SECP256K1_BUILD',
+    'USE_ASM_ARM',
+    'USE_NUM_NONE',
+    'USE_FIELD_INV_BUILTIN',
+    'USE_SCALAR_INV_BUILTIN',
+    'USE_FIELD_10X26',
+    'USE_SCALAR_8X32',
+    'USE_ECMULT_STATIC_PRECOMPUTATION',
+    'USE_EXTERNAL_DEFAULT_CALLBACKS',
+    ('ECMULT_WINDOW_SIZE', '8'),
+    'ENABLE_MODULE_GENERATOR',
+    'ENABLE_MODULE_RANGEPROOF',
+    'ENABLE_MODULE_RECOVERY',
+    'ENABLE_MODULE_ECDH',
+]
+SOURCE_MOD_SECP256K1_ZKP += [
+    'vendor/secp256k1-zkp/src/secp256k1.c',
 ]
 
 # modtrezorio
@@ -404,11 +431,33 @@ source_mpyc = env.FrozenCFile(
 env.Depends(source_mpyc, qstr_generated)
 
 #
+# static secp256-zkp ecmult context
+#
+
+host_env = Environment(ENV=os.environ)
+host_env.Replace(
+    CC=os.getenv('CC_FOR_BUILD') or 'gcc',
+    COPT='-O2',
+    CPPPATH='vendor/secp256k1-zkp',
+)
+gen_context = host_env.Program(
+    target='vendor/secp256k1-zkp/gen_context',
+    source='vendor/secp256k1-zkp/src/gen_context.c',
+)
+
+secp256k1_zkp_ecmult_static_context = host_env.Command(
+    target='vendor/secp256k1-zkp/src/ecmult_static_context.h',
+    source='vendor/secp256k1-zkp/gen_context',
+    action='cd ${SOURCE.dir}; ./gen_context',
+)
+
+#
 # Program objects
 #
 
 obj_program = []
 obj_program.extend(env.Object(source=SOURCE_MOD))
+obj_program.extend(env.Object(source=SOURCE_MOD_SECP256K1_ZKP, CCFLAGS='$CCFLAGS -Wno-unused-function'))
 obj_program.extend(env.Object(source=SOURCE_FIRMWARE))
 obj_program.extend(env.Object(source=SOURCE_MICROPYTHON))
 obj_program.extend(env.Object(source=SOURCE_MICROPYTHON_SPEED, COPT='-O3'))
@@ -435,6 +484,7 @@ obj_program.extend(
         ' $SOURCE $TARGET', ))
 
 env.Depends(obj_program, qstr_generated)
+env.Depends(obj_program, secp256k1_zkp_ecmult_static_context)
 
 program_elf = env.Command(
     target='firmware.elf',
--- a/embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
+++ b/embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
@@ -1,0 +1,306 @@
+/*
+ * This file is part of the TREZOR project, https://trezor.io/
+ *
+ * Copyright (c) SatoshiLabs
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "common.h"
+#include "py/objstr.h"
+
+#include "vendor/secp256k1-zkp/include/secp256k1.h"
+#include "vendor/secp256k1-zkp/include/secp256k1_ecdh.h"
+#include "vendor/secp256k1-zkp/include/secp256k1_preallocated.h"
+#include "vendor/secp256k1-zkp/include/secp256k1_recovery.h"
+
+// The minimum buffer size can vary in future secp256k1-zkp revisions.
+// It can always be determined by a call to
+// secp256k1_context_preallocated_size(...) as below.
+STATIC uint8_t g_buffer[(1UL << (ECMULT_WINDOW_SIZE + 4)) + 208] = {0};
+
+void secp256k1_default_illegal_callback_fn(const char *str, void *data) {
+  (void)data;
+  mp_raise_ValueError(str);
+  return;
+}
+
+void secp256k1_default_error_callback_fn(const char *str, void *data) {
+  (void)data;
+  __fatal_error(NULL, str, __FILE__, __LINE__, __func__);
+  return;
+}
+
+STATIC const secp256k1_context *mod_trezorcrypto_secp256k1_context(void) {
+  static secp256k1_context *ctx;
+  if (ctx == NULL) {
+    size_t sz = secp256k1_context_preallocated_size(SECP256K1_CONTEXT_SIGN |
+                                                    SECP256K1_CONTEXT_VERIFY);
+    if (sz > sizeof g_buffer) {
+      mp_raise_ValueError("secp256k1 context is too large");
+    }
+    void *buf = (void *)g_buffer;
+    ctx = secp256k1_context_preallocated_create(
+        buf, SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
+
+    uint8_t rand[32];
+    random_buffer(rand, 32);
+    int ret = secp256k1_context_randomize(ctx, rand);
+    if (ret != 1) {
+      mp_raise_msg(&mp_type_RuntimeError, "secp256k1_context_randomize failed");
+    }
+  }
+  return ctx;
+}
+
+/// def generate_secret() -> bytes:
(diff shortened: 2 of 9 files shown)


embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
```diff
@@ -0,0 +1,247 @@
...
+
+#include "py/objstr.h"
+
+#include "ecdsa.h"
+#include "secp256k1.h"
+#include "secp256k1_ecdh.h"
+#include "secp256k1_preallocated.h"
+#include "secp256k1_recovery.h"
+
+STATIC uint8_t g_buffer[4500] = {0};
```
real-or-random: We should decrease this to the right number. I think it's actually somewhere around 4200 bytes, so we can save a few hundred bytes.

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
```diff
@@ -0,0 +1,247 @@
...
+    (void) data;
+    mp_raise_ValueError(str);
+    return;
+}
+
+void secp256k1_default_error_callback_fn(const char* str, void* data) {
+    (void) data;
+    mp_raise_msg(&mp_type_RuntimeError, str);
+    return;
+}
```
real-or-random: These are mentioned error callbacks.

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
```diff
@@ -0,0 +1,247 @@
...
+    recid &= 3;
+
+    secp256k1_ecdsa_recoverable_signature ec_sig;
+    if (!secp256k1_ecdsa_recoverable_signature_parse_compact(ctx, &ec_sig, (const uint8_t *)sig.buf + 1, recid)) {
+        mp_raise_ValueError("Invalid signature");
+    }
+    secp256k1_pubkey pk;
+    if (!secp256k1_ecdsa_recover(ctx, &pk, &ec_sig, (const uint8_t *)dig.buf)) {
+        // Copying "return None" behaviour from secp256k1 module; but why do
+        // we do this rather than raising an exception?
```
real-or-random: Looks like a TODO.

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
```diff
@@ -0,0 +1,247 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "py/objstr.h"
+
+#include "ecdsa.h"
+#include "secp256k1.h"
```
real-or-random: This is the ambiguous #include.

real-or-random:
Note to self: We should also call `secp256k1_context_randomize` (https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L641) after creating the context

prusnak:
1) I would prefer not changing `import secp256k1` to `import secp256k1_zkp` in this pull request. It simplifies reviewing and also I am not sure we are going to switch immediately. It makes sense to use `secp256k1_zkp` for Liquid, but I am not in favor of using it everywhere unless there are proper benchmarks done which prove the zkp implementation is faster and/or has other big advantages.

real-or-random:
>  I would prefer not changing `import secp256k1` to `import secp256k1_zkp` in our Python code via this pull request. It simplifies reviewing and I am not sure whether we are going to switch immediately. It makes sense to use `secp256k1_zkp` for Liquid, but I am not in favor of using it everywhere unless there are benchmarks done which prove the zkp implementation is faster and/or has other big advantages.

No problem, then we should first focus on just getting zkp imported and included in the build. I'll update this PR then. Anyway, it was still good to have these other commits to demonstrate that the build worked and can actually be used, and for the future.

>  I agree we should use directories in `#include` to avoid clashes. I will address this in another pull request.

Okay, I can also just do it when I update this.

prusnak:
> Anyway, it was still good to have these other commits to demonstrate that the build worked and can actually be used, and for the future.

Sure, I 100% agree. To be clear, I am not opposed to using secp256k1_zkp globally for ECDSA/ECDH operations. But let's test that first.

SConscript.firmware
```diff
@@ -85,6 +86,31 @@ SOURCE_MOD += [
     'vendor/trezor-crypto/sha3.c',
 ]
 
+# libsecp256k1-zkp
+CPPPATH_MOD += [
+    'vendor/secp256k1-zkp/',
```
prusnak: Do we need to have this line in the list? (`'vendor/secp256k1-zkp/'`)

prusnak:
Maybe we can leverage the fact the tests for `secp256k1` and `secp256k1_zkp` ar the same.
Something like this comes to mind:

```
from trezor.crypto.curve import secp256k1, secp256k1_zkp

class Secp256k1Common(unittest.TestCase):
   def test_foo(self):
       # use impl.method instead of secp256k1.method

class TestCryptoSecp256k1(Secp256k1Common): 
    def __init__(self):
         self.impl = secp256k1

class TestCryptoSecp256k1Zkp(Secp256k1Common): 
    def __init__(self):
         self.impl = secp256k1_zkp
```

SConscript.firmware
real-or-random: Yes, that's how library is built. Otherwise you'll get 
```
vendor/secp256k1-zkp/src/secp256k1.c:7:10: fatal error: include/secp256k1.h: No such file or directory #include "include/secp256k1.h"
```

Is this problematic?

prusnak: It's not. OK

romanz:
A small update to the PR (following a discussion with @real-or-random):
I have replaced https://github.com/trezor/trezor-core/commit/f8b5bbc17f70be4333666fd8421b47c0dfa3cc97 by https://github.com/trezor/trezor-core/commit/55a39486ce9a06a0251c329aa036d34a551f77ba (moving the read-only precomputed tables from the `.flash` to `.flash2` section, which AFAIU is not used for storing executable code).
The resulting binary should compile and run successfully on TREZOR model T.

prusnak:
> moving the read-only precomputed tables from the .flash to .flash2 section, which AFAIU is not used for storing executable code

Good!

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
real-or-random: done

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
real-or-random: fixed

real-or-random:
I added two commits to address the file name issues and the size of the context.

@prusnak Do the callbacks look right to you?

prusnak:
@real-or-random offtopic: How hard is to add NIST P-256 curve to secp256k1 library? Is this even possible or desired?

apoelstra:
@prusnak it would basically mean rewriting the library. It is neither possible nor desired :)

prusnak:
Got it :)

real-or-random:
We have now https://github.com/ElementsProject/secp256k1-zkp/pull/53 which (when merged) removes the need to use a modified branch/version of secp256k-zkp.

The remaining issues are the top three items in the list in first comment here in this PR
@prusnak Can you comment on these? :)

prusnak:
> I replaced them by functions which throw Python exceptions. Is this the right thing to do on trezor?

Yes

> 4096 bytes (of RAM). Is this is fine for you as memory footprint? 

Yes

> The table(s) are stored in a secp256k1_context which is currently created when it's first used. Is that okay or do we want to create at program startup?

I think that's fine as a PoC.

prusnak:
Can you please squash the changes into a single commit and resolve conflicts? We have reformatted our code using clang-format (config is in `.clang-format` but it should be picked up automatically, just use `clang-format -i` on newly added C files).

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
real-or-random: Yes, this is still open. We raise an exception for every other failure mode, so I see no reason why we should not raise an exception here. Any ideas?

real-or-random:
Okay, I squashed and changed it to remove the commits that change the apps.
Two issues open left:
  * Your TODO where we don't raise an exception, see above.
  * We need to wait until  https://github.com/ElementsProject/secp256k1-zkp/pull/53 is merged and change the submodule according.

real-or-random:
> > The table(s) are stored in a secp256k1_context which is currently created when it's first used. Is that okay or do we want to create at program startup?
> 
> I think that's fine as a PoC.

Yes, indeed. I think in a PR that actually changes apps to use secp256k1-zkp, it's best to change this to a model where the verification table is only created when it's necessary (which is rare). Currently, we would create the verification table even when we would call into secp256k1-zkp for signing, which is kind of silly because signing is much more common and we don't need to build a table for signing (because the signing table it's statically built).

embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h
real-or-random: Well, so the documentation of the function says that it should return `None` if the signature is invalid, so `None` makes sense. However, the actual behavior of the function is that either returns `None` or raises `ValueError` if the signature is invalid.

This inconsistency lead to a bug (#422), which was  fixed in https://github.com/trezor/trezor-core/commit/d5fb2a477a7267ec6c88b7e09c39c8c9a63022fc by handling both `None` and an exceptions equally in the caller code. (The Ethereum app is the only caller.)

I think this can be improved by making the function consistently return `None` no matter what the failure is, but in a different PR. I will remove your comment then at least here. If you agree with this conclusion, I'm volunteering to create that other PR, it should just take a few minutes.

real-or-random:
I added two commits (will squash again if you approve):
  * I changed my mind about one the callbacks, namely the `error_callback`. This is only triggered when some internal check fails and we really just want to crash the entire device, because continuing execution could mean undefined behavior or leaking secrets etc, see https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/include/secp256k1.h#L247 for an explanation. So this is a last resort thing, and I changed it to call `__fatal_error()` instead.  
  * I removed your comment about returning None as explained here https://github.com/trezor/trezor-core/pull/500#discussion_r271510331.

With these two commits, this PR is ready to be merged except that we need to wait for #53 to be merged and adjust the submodule.

prusnak:
>  (will squash again if you approve):

Agreed. Please squash.

> this PR is ready to be merged except that we need to wait for #53 to be merged and adjust the submodule.

Ack. we'll review it in the meantime. @jpochyla can you have a look, please?

real-or-random:
Squashed.

real-or-random:
Okay, sorry, I pushed again because I additionally enabled some ASM optimizations for ASM in secp256k1-zkp.

prusnak:
Okay, we are ready to merge. We'll wait until you merge https://github.com/ElementsProject/secp256k1-zkp/pull/53 and adjust a submodule in this PR.

real-or-random:
Okay force-pushed, I touched only `.gitmodules`. The other PR is not merged yet but should be today.

apoelstra:
Merged the other PR.

prusnak:
@real-or-random do you want to change the revision of the included submodule or should we merge as is?

real-or-random:
Oh yes I should have done this. I'll update it later. :)

real-or-random:
fixed

prusnak:
Great work @real-or-random, thanks for the contribution!

PR #503, opened 2019-03-18
cardano: taint pages to prevent blank screen

### LINKED ISSUE #497: Swiping up and down when signing cardano transaction leads to blank screen
I was testing sending of a cardano transaction on the latest released firmware (2.1.0) and the screen where the amount and output address is being confirmed empties when I swipe up and back down (it's paginated). Only the reject/confirm buttons stay in place. You can reproduce it here: http://adalite.io/ If you need ADA I can send some to you.

I'm not sure if we are doing something wrong with the pagination: https://github.com/trezor/trezor-core/blob/20c97e85addb73a2e8e773a3c461319925c7833f/src/apps/cardano/layout/__init__.py#L37

j-matus:
Taint pages to ensure the texts are rendered when swiping. Closes #497

### DIFF
--- a/src/apps/cardano/layout/__init__.py
+++ b/src/apps/cardano/layout/__init__.py
@@ -66,6 +66,7 @@ def create_renderer(confirmation_wrapper):
         if page >= page_count:
             page = page_count - 1
 
+        pages[page].taint()
         content = Scrollpage(pages[page], page, page_count)
         if page + 1 >= page_count:
             return await confirmation_wrapper(content)


refi93:
@jpochyla btw, is this "tainting" the recommended way to deal with refreshing paginated content?

tsusanka:
@jpochyla let's merge this?

tsusanka:
I've tested this PR and it fixes the issue.

PR #505, opened 2019-03-20
Added KIN

chancity:
Added stellar based coin KIN

https://www.kin.org/


prusnak:
There is clearly a better way than just copy-pasting stellar code to another location. Think more about how you can reuse the Stellar messages to implement your coin.

PR #518, opened 2019-03-26
Tezos: add voting support

adonagy:
Added support for signing tezos voting operations - proposal and ballot. With these operations we allow tezos user to participate in the protocol amendment process with a Trezor T device


adonagy:
Depends on https://github.com/trezor/trezor-common/pull/300

tsusanka:
Just a side note: let's make sure UI is correct - not just visually but also from the functioning side (Cancel buttons) etc.

adonagy:
@tsusanka The UI is correct functionally as well. When we sign a ballot operation,  there is the cancel button from `request_confirm` in case the user decides to cancel the operation. 

The proposal screen also contains the same cancel button. When we are signing multiple proposals in the same operation, we use `Scrollpage`, with every proposal hash shown on a different page, the cancel button shown on the last screen.

src/apps/tezos/layout.py
```diff
@@ -1,5 +1,9 @@
-from trezor import ui
+from micropython import const
```
matejcik: you don't seem to use this

src/apps/tezos/layout.py
```diff
@@ -66,6 +70,51 @@ def split_address(address):
...
+async def require_confirm_ballot(ctx, proposal, ballot):
+    text = Text("Submit ballot", ui.ICON_SEND, icon_color=ui.PURPLE)
+    text.bold("Ballot: {}".format(ballot))
+    text.bold("Proposal:")
+    text.mono(*split_proposal(proposal[0]))
+    await require_confirm(ctx, text, ButtonRequestType.SignTx)
+
+
+# use, when there are more then one proposals in one operation
+async def show_proposals(ctx, proposals):
```
matejcik: please merge this with `require_confirm_proposal` (and rename to proposal*s*, probably), instead of making the decision in `sign_tx`

matejcik: also the paginator is smart enough to show only one page when required. you can add a `title` argument to `show_proposal_page` and supply the "proposal/proposals" string this way.

src/apps/tezos/writers.py
```diff
@@ -0,0 +1,24 @@
+from apps.common.writers import (
```
matejcik: as noted in the baking PR, this file should not exist.

please use the `write_uint16_be` implementation from your baking PR, and move `write_bool` to helpers
in fact, this is almost certainly going in sooner than baking support, so you can then use this as the basis of the baking PR

src/apps/tezos/helpers.py
```diff
@@ -22,6 +22,8 @@
     "edsig": [9, 245, 205, 134, 18],
     # operation hash
     "o": [5, 116],
+    # protocola hash
```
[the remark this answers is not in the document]
adonagy: Yes, typo

src/apps/tezos/layout.py
adonagy: Cool, will do that.

adonagy: I looked at the paginator again, and it indeed shows the last page, but it also allows to swipe up. 
So when the user accidentally swipes up a the device raises a FrimwareError, because there is no other page to swipe to. If I recall correctly, this was the reason behind using another screen for a single proposal. 

ISTM the `ui.scroll.change_page` is not handling this edge case, right?

matejcik: that's a bug, but we'll fix it in `change_page`

prusnak:
Superseded by https://github.com/trezor/trezor-firmware/pull/41

PR #527, opened 2019-04-02
Permanent display rotation

### LINKED ISSUE #162: Display rotation in lock screen & make it persistent
Observations:
a) People sometimes thinks that display rotation is a software bug.
b) Lock screen and PIN entry is not affected by the rotation settings, which defeats its purpose for heads-down mounts.
c) Settings currently doesn't survive device restart.

Solution:
1) Make it as a device settings over ApplySettings() message as unprotected value (similar to background image & label).
2) Make PIN entry and lock screen compatible with rotation.
3) Disable rotation over swipe gesture.

tsusanka:
Permanently store display rotation and disable swipe gesture, closes #162. After merged in trezor-common I'll force push here.

common: https://github.com/trezor/trezor-common/pull/304
python-trezor: https://github.com/trezor/python-trezor/pull/372

Any ideas how to test this?


src/apps/homescreen/homescreen.py
```diff
@@ -1,15 +1,12 @@
...
 
 async def homescreen():
-    while True:
-        await ui.backlight_slide(ui.BACKLIGHT_DIM)
-        display_homescreen()
-        await ui.backlight_slide(ui.BACKLIGHT_NORMAL)
-        await swipe_to_rotate()
+    await ui.backlight_slide(ui.BACKLIGHT_DIM)
+    display_homescreen()
+    await ui.backlight_slide(ui.BACKLIGHT_NORMAL)
```
tsusanka: This is probably incorrect, because I've found out it breakes the ability to do screenshots (and probably other things?). However, I'm not sure how to fix this. Should it stay in the loop? With a blocking call at the end? But waiting for what?

prusnak:
trezor/trezor-common#304 merged

src/apps/homescreen/homescreen.py
prusnak: // cc @jpochyla

tsusanka:
Updated

prusnak:
Merged via bd812fecf9e0ca4573f15f62bec67d6e6f4875e5

PR #531, opened 2019-04-02
Save screen after every refresh

jpochyla:
If environment variable `TREZOR_SAVE_SCREEN` is set, every screen refresh produces a screenshot in current directory. Useful for integration testing and generating workflow documentation.


tsusanka:
Cool! Tested ACK.

PR #532, opened 2019-04-03
Pass Keychain to path validation function to throw error before warning

### LINKED ISSUE #519: Keychain.derive and path validations
Currently, we first call `paths.validate_path` and prompt the user in case it is a non-standard path. If so and the user agrees it may happen that the Keychain throws `failure: Forbidden key path` if it is outside the app's namespace, which is quite unfriendly.

Do we want to deal with this? Maybe just swap those two operations?

---

You can reproduce this using `trezorctl lisk_get_address -n "43"`. [GIF](https://user-images.githubusercontent.com/1835345/55010263-32818e80-4fe4-11e9-81d3-feebf6a569ad.gif)

tsusanka:
Pass Keychain to path validation function, which then calls new function `validate_path` on the Keychain to see if it allows this path. This closes #519 .

For the time being, this ignores curves whatsoever. So, if the user provides path with a non-hardened item for an app, which uses uncompatible curve (ed25519), we still show a warning and then fail, which is exactly what we want to solve. Example: sending `44'/1729'/10'/11` to Tezos. We could either:

1. Ignore this. It is somewhat an edge case.
1. Pass curve to the validate_path function. It is yet another parameter (which is a bit ugly), but we could check the case above and also if the curve matches one of the Keychain's.
1. In Keychain.validate_path check if the curve is ed25519 and in that case forbid non-hardened items.

### DIFF
--- a/src/apps/cardano/get_address.py
+++ b/src/apps/cardano/get_address.py
@@ -10,7 +10,7 @@ from apps.common.layout import address_n_to_str, show_address, show_qr
 async def get_address(ctx, msg):
     keychain = await seed.get_keychain(ctx)
 
-    await paths.validate_path(ctx, validate_full_path, path=msg.address_n)
+    await paths.validate_path(ctx, validate_full_path, keychain, msg.address_n)
 
     try:
         address, _ = derive_address_and_node(keychain, msg.address_n)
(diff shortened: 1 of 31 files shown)


tsusanka:
I'm merging this and I'll file another PR for the _curve_ issue.

PR #535, opened 2019-04-04
extmod/modtrezorcrypto: return False or None consistently when a signature verification fails

real-or-random:
So far, we either return False (or None for public recovery) or raise a
ValueError (e.g., when the length of the signature). This is
inconsistent and dangerous because the inputs to signature verification
may be attacker-provided and cannot be assumed to be well-formed.
    
This led to issue #422 where a firmware error is raised when an invalid
signature is is provided. This has been fixed for the ethereum app but
not for the wallet app. This commit addresses the problem at the core of
the issue, i.e., at the verification functions in extmod such that all
apps are covered.

This is a draft PR because it builds on top of #500. It should not be merged currently. But the only commit is the last one, and this is ready for review.

### DIFF
--- a/embed/extmod/modtrezorcrypto/modtrezorcrypto-ed25519.h
+++ b/embed/extmod/modtrezorcrypto/modtrezorcrypto-ed25519.h
@@ -147,13 +147,13 @@ STATIC mp_obj_t mod_trezorcrypto_ed25519_verify(mp_obj_t public_key,
   mp_get_buffer_raise(signature, &sig, MP_BUFFER_READ);
   mp_get_buffer_raise(message, &msg, MP_BUFFER_READ);
   if (pk.len != 32) {
-    mp_raise_ValueError("Invalid length of public key");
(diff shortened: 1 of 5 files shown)


real-or-random:
rebased on #500

prusnak:
@real-or-random I guess this one should be switched to regular PR? (After https://github.com/trezor/trezor-core/pull/500 has been merged)

real-or-random:
Yes, I had left this as draft because #500 is not merged yet, but this is ready for review!

real-or-random:
Okay, sure, there were changes to master of course and I was confused. It's properly rebased now.

tsusanka:
ACKed by @jpochyla. We're moving to monorepo: https://github.com/trezor/trezor-firmware . I've dared to merge it to master directly, so you don't have to resubmit. https://github.com/trezor/trezor-firmware/commit/5dc0a1ee0a8023fec2d385c65a63f8cbeb40d3a3

PR #539, opened 2019-04-08
Add Onegram support

Adman:


Adman:
Note: I will link the common repo back when the review is done or the common is merged.

prusnak:
Let's not add this insignificant coin. Feel free to create a fork of the firmware and deliver it to Onegram users.

tsusanka:
For reference to other users: please read this https://wiki.trezor.io/Developers_guide:Contributing before doing some larger amount of work

Issue #506, opened 2019-03-20
Display Custom Locktime On Trezor Screen for BTC


yuriygeorge:
I'm a Trezor app developer and for my application, I need to be able to set the locktime parameter, i.e., so the transaction is signed by Trezor and is only valid when a specific block number is reached. However, I notice that the custom "locktime" field is not displayed on the Trezor display screen when set. I understand that when it is not set, no need to display it. But for my application when I need to set the locktime, how can the user be certain that the locktime field was actually set? Is it possible in the near future to make Trezor display the locktime field if it is set? Otherwise, seems to defeat the purpose if the user isn't able to verify it on the trezor device.

tsusanka:
TLDR: if locktime is different than the default value add a screen for confirming it

ciny:
Based on discussion in meeting: we'll be displaying raw value and differentiate between blockheight/timestamp

locktime:
blockheight: 571032

or

locktime:
timestamp: 500 000 001

tsusanka:
So, @prusnak should we consider 0xFFFFFFFF and 0xFFFFFFFE as allowed values without confirmation? I believe so

prusnak:
@tsusanka I got that mixed. 0xFFFFFFFF and 0xFFFFFFFE are nSequence values of the input. Transaction is RBF when all inputs have nSequence == 0xFFFFFFFE. No work is need to be done.

yuriygeorge:
Thanks!

PR #540, opened 2019-04-08
#506 display custom locktime

ciny:
Dates need work on embed platforms. All tests pass.

### DIFF
--- a/src/apps/wallet/sign_tx/layout.py
+++ b/src/apps/wallet/sign_tx/layout.py
@@ -8,9 +8,15 @@ from trezor.utils import chunks, format_amount
 from apps.common.confirm import confirm, hold_to_confirm
 from apps.wallet.sign_tx import addresses, omni
 
+# from datetime import datetime
+
 
 def format_coin_amount(amount, coin):
     return "%s %s" % (format_amount(amount, 8), coin.coin_shortcut)
+
+
+# def format_locktime(lock_time):
+#     return datetime.utcfromtimestamp(lock_time).strftime('%Y-%m-%d %H:%M:%S')
 
 
 def split_address(address):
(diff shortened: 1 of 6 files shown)


src/apps/wallet/sign_tx/layout.py
```diff
@@ -65,3 +84,16 @@ def split_op_return(data):
     text = Text("Confirm sending", ui.ICON_SEND, icon_color=ui.RED)
     text.normal("Trying to spend", "coins from another chain.", "Continue?")
     return await confirm(ctx, text, ButtonRequestType.SignTx)
+
+
+async def confirm_nondefault_locktime(ctx, lock_time):
+    text = Text("Confirm locktime", ui.ICON_SEND, icon_color=ui.GREEN)
+    if lock_time < 500000000:
```
prusnak: Let's extract `500000000` as a constant similarly as `_EMBEDDED_TIMESTAMP_DIFF` is extracted.

src/apps/wallet/sign_tx/layout.py
```diff
@@ -8,11 +11,27 @@
...
+def convert_to_embedded_timestamp(unix_timestamp):
+    return max(0, unix_timestamp - _EMBEDDED_TIMESTAMP_DIFF)
+
+
+# dates before 1.1.2000 will resolve to 1.1.2000 on non-unix
+def format_date(timestamp):
+    if not sys.platform.equals("unix"):
+        timestamp = convert_to_embedded_timestamp(timestamp)
+
+    return utime.localtime(timestamp)
```
prusnak: We might consider avoiding using the `utime` module completely and compute the date in pure Python instead.

src/apps/wallet/sign_tx/layout.py
ciny: done

src/apps/wallet/sign_tx/layout.py
ciny: based on what we agreed in meeting all timestamp->formatted date functionality was removed.

prusnak:
Great, thanks!

PR #541, opened 2019-04-08
tests: monero binary test added

ph4r05:
Testing with the official Monero binary client and trezor tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ build/
 build-docker/
 emu.config
 compile_commands.json
+tests/trezor_monero_tests{,.log}
--- a/tests/run_tests_device_emu_monero.sh
+++ b/tests/run_tests_device_emu_monero.sh
@@ -1,27 +1,96 @@
 #!/bin/bash
 
-MICROPYTHON=../build/unix/micropython
+: "${RUN_PYTHON_TESTS:=0}"
+: "${FORCE_DOCKER_USE:=0}"
+
+SDIR="$(SHELL_SESSION_FILE= && cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+CORE_DIR="$SDIR/.."
+MICROPYTHON=$CORE_DIR/build/unix/micropython
 PYOPT=0
 
 # run emulator
-cd ../src
+cd $CORE_DIR/src
 $MICROPYTHON -O$PYOPT main.py >/dev/null &
 upy_pid=$!
+cd -
 sleep 1
 
 export TREZOR_PATH=udp:127.0.0.1:21324
+DOCKER_ID=""
+
+# Test termination trap
+terminate_test() {
+  if [[ $# > 0 ]]; then error=$1; fi
+  kill $upy_pid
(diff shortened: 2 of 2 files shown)


ph4r05:
I've just tested that the binary compiled for Ubuntu 14.04 works also for Ubuntu {16.04, 18.04}. 

Do you want to somehow condition binary on the current OS? Do you start these tests also on OSX?

tsusanka:
I've tested this on my computer and it works, I have a few suggestions:

1. rename trezor_tests to trezor_monero_tests or similar
1. add trezor_tests{,.log} to .gitignore
1. fetch a hash first and compare it with trezor_tests file if present, so we can run the tests repeatedly without having to download the binary again

----

@jpochyla What's your take on OSX? Do we want to maintain that or we're fine with linux only? I'm not sure myself

ph4r05:
> * rename trezor_tests to trezor_monero_tests or similar
> * add trezor_tests{,.log} to .gitignore
> * fetch a hash first and compare it with trezor_tests file if present, so we can run the tests repeatedly without having to download the binary again

I've addressed the suggestions, pls check.

For OSX I am creating an ephemeral docker container running the tests. Due to limitation of the UDP transport I had to use `--network=host` for the container and `socat` to make emulator accessible on localhostfrom within the container (monero client can connect only to localhost emulators)

tsusanka:
> I've addressed the suggestions, pls check.

I've pushed some typos. Also, the `pwd` command gets evaluated to the root of the directory, so the .gitignore does not seem to work. I'm not really sure why though because I am in the tests directory.

---

From the docker side, probably someone with OSX should try it. @jpochyla want to give it a go?

tsusanka:
I couldn't push to your PR, can you cherry-pick please? https://github.com/trezor/trezor-core/commit/ed09b6b8d94561e56973cb661ec663c8a6651b70

ph4r05:
> I've pushed some typos. Also, the `pwd` command gets evaluated to the root of the directory, so the .gitignore does not seem to work. I'm not really sure why though because I am in the tests directory.

I was thinking about it also, the script does `cd ..` which is quite subtle and hard to spot. I've taken it from `run_tests_device_emu.sh`.

> I couldn't push to your PR, can you cherry-pick please? ed09b6b 

Done

ph4r05:
> From the docker side, probably someone with OSX should try it. @jpochyla want to give it a go?

I've made a slight change, you can test the docker as well now:

```bash
RUN_PYTHON_TESTS=0 FORCE_DOCKER_USE=1 ./run_tests_device_emu_monero.sh
```

ph4r05:
Note: I will update the binary in the evening so pls don't merge until then, thanks

jpochyla:
Works well, thanks!

.gitignore
```diff
@@ -9,3 +9,5 @@ build/
 build-docker/
 emu.config
 compile_commands.json
+trezor_monero_tests{,.log}
+tests/trezor_monero_tests{,.log}
```
jpochyla: Might be a good idea to use file-relative location for the download directory, not cwd-relative one, so we don't need both of there here.

ph4r05: True that. The script is now cwd independent.

PR #542, opened 2019-04-08
paths: validate curve as well

tsusanka:
As discussed in https://github.com/trezor/trezor-core/pull/532, this PR adds also curve checking to the existing path validation.


matejcik:
from a cursory read, this would break Cardano's non-hardened derivation on ed25519

(also not catching NEM's ed25519-keccak)

Also passing the curve argument seems kind of pointless? The curve is specified in the namespace, and right now it is not user-configurable, so we're validating whether you're passing the same constant to `validate_path` that you also passed to namespace in the next file over.

A more sensible argument would be something like `allow_unhardened=None|True|False`, with:
* `False` explicitly disallowing (for NEM which doesn't use `"ed25519"`)
* `True` for explicitly allowing (for Cardano which can do unhardened derivation on ed25519)
* and `None` for default (disallow for `"ed25519"`, allow otherwise; or more generally, allow for curves that can do it and disallow for those that can't

tsusanka:
> from a cursory read, this would break Cardano's non-hardened derivation on ed25519

It shouldn't, Cardano has its own [Keychain](https://github.com/trezor/trezor-core/blob/master/src/apps/cardano/seed.py).

> (also not catching NEM's ed25519-keccak)

Great point!

-----

Yeah, maybe you're right. I was also thinking if we shouldn't do this somehow automagically. For example, the validate_path function would know where from it was called, so it can fetch the app's CURVE constant from the module's `__init__`. But it seems as too much woodoo.

PR #543, opened 2019-04-08
Update run_tests.sh: show totals, allow interrupts, add colors

MrMebelMan:
* show the number of passed and failed tests

* allow the user to stop running tests via CTRL+C

* colorize the output


tsusanka:
This is great! Thanks.

PR #549, opened 2019-04-11
utils: optimize __slots__ comparison

MrMebelMan:
I've noticed that when comparing `__slots__`' attributes, a new dictionary is being created.
To benefit from improved speed and memory usage we should access these attributes directly :muscle:

### DIFF
--- a/src/trezor/utils.py
+++ b/src/trezor/utils.py
@@ -104,10 +104,14 @@ def obj_eq(l, r):
     """
     if l.__class__ is not r.__class__:
         return False
-    if hasattr(l, "__slots__"):
-        return obj_slots_dict(l) == obj_slots_dict(r)
-    else:
+    if not hasattr(l, "__slots__"):
         return l.__dict__ == r.__dict__
+    if l.__slots__ is not r.__slots__:
+        return False
+    for slot in l.__slots__:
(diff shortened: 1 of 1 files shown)


src/trezor/utils.py
```diff
@@ -104,28 +104,20 @@ def obj_eq(l, r):
...
     if l.__class__ is not r.__class__:
         return False
-    if hasattr(l, "__slots__"):
-        return obj_slots_dict(l) == obj_slots_dict(r)
-    else:
+    if not hasattr(l, "__slots__"):
         return l.__dict__ == r.__dict__
+    for slot in l.__slots__:
+        if getattr(l, slot, None) != getattr(r, slot, None):
+            return False
```
jpochyla: This doesn't check that `r` has some slots that `l` does not have.

matejcik: technically `l.__slots__ is r.__slots__`, no?

jpochyla: Yep! 👍

MrMebelMan: > technically `l.__slots__ is r.__slots__`, no?

fixed, thanks!

prusnak:
:tada:

PR #552, opened 2019-04-18
xmr: live refresh progress starts from 0

ph4r05:
Minor fix on live refresh progress monitoring, it started from -1 before.


tsusanka:
This was used only in UI so it does not have any other consequences, right?

Merged via https://github.com/trezor/trezor-firmware/commit/021a8a0cb86fa44c673797661876749e67c9ebb3. From now on please open all PRs in trezor-firmware.

